PR #143, opened 2015-01-26
fix tab key inserting hard tab instead of spaces

Gibstick:
Style guide says spaces instead of hard tab. I've mentioned in an email but it doesn't seem to have been fixed yet. HTH :) Another option is to bind Tab to indentAuto, as this will mimic DrRacket's behaviour with Tab, but I don't think this is typical editor behavior.


e45lee:
Great!  Thanks for your PR.  As a heads up, this won't immediately hit the Seashell that is currently available to students, as we're in the middle of a frontend cleanup, but you should see the improvements in the next few weeks.

Issue #142, opened 2015-01-22
Implement git features: commit, revert


kpalway:
This has been requested by students several times. I can't recall if we decided this would be too confusing or not.

kcolford:
If you can at least enable the commit feature, then you can add in a help message describing it as "Saving a Snapshot of your current project, so that if something goes wrong then the staff [or you] can can restore it to the way it was before.  Leave a descriptive message so that when you go back, you know which backup is the one you want."

Also, the commit feature is very important because otherwise you leave all edits as uncommited changes in the repository, which means you can't fix (and therefore can't realistically distribute) object files from the skeleton repository for an assignment.  This is becuase git-pull uses git-merge, and from the man page of git-merge(1):

>**Warning**: Running git merge with non-trivial uncommitted changes is
>discouraged: while possible, it may leave you in a state that is hard
>to back out of in the case of a conflict.

Which means that if the user has edited any of the files in a way that would cause a conflict, you can't just use the merge strategy "ours" to keep the user code there since it isn't even in the index to begin with.

I'm putting together a patch right now for this, it seems you already have most of the code needed to make it happen, so I'll send a pull request soon.

e45lee:
It's probably a sensible thing to also commit on/before each submission to Marmoset.

e45lee:
We have decided to not use git after all.  I am closing this issue to reflect this.

PR #144, opened 2015-01-31
[frontend] Added "commit" button according to issue #142.

kcolford:
The dialog button is also bound to the key control+S to conform to the
style of the other buttons.  It was placed to the right of the TEST
button because like the TEST and RUN buttons it affects the project as
a whole (not just a specific file).

### DIFF
--- a/src/frontend/frontend/frontend.js
+++ b/src/frontend/frontend/frontend.js
@@ -87,6 +87,23 @@ angular.module('frontend-app', ['seashell-websocket', 'seashell-projects', 'jque
           }).result;
         };
       }])
+  // Commit Project Modal Service
+  .factory('CommitProjectModal', ['$modal', 'error-service',
+      function ($modal, errors) {
+        return function (project) {
+          return $modal.open({
+            templateUrl: "frontend/templates/commit-project-template.html",
+            controller: ['$scope', 'socket', 'error-service',
+            function ($scope, ws, errors) {
+              $scope.commit_descr = "";
+              $scope.CommitProject = function () {
+                $scope.$close();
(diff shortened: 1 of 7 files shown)


e45lee:
Unfortuantely, we can't merge your pull-request as is; we are currently refactoring the Seashell frontend code, so your PR won't work right now.

In particular, we're using Angular these days.  There's a few examples of how modals are set up in
the Seashell code w/ Angular (see src/frontend/frontend/frontend.js).  If you could rewrite the dialog
handling code that you currently have (in src/frontend/js/setup.js and src/frontend/js/ui/menu.js),
then I can go ahead and try to merge this PR.

kcolford:
Angular has been incorperated, I don't know how you are binding UI events to API calls, so that will have to be done by someone else or someone will have to point out where this is being done already.

I left the old non-Angular code in because it can still be backported to the old stable release.

e45lee:
A good example of how UI events (and data) are being bound to API calls can be found at https://github.com/cs136/seashell/blob/master/src/frontend/frontend/frontend.js#L66-L89, with the template found
https://github.com/cs136/seashell/blob/master/src/frontend/frontend/templates/new-file-template.html.

I would prefer if you would set up the commit dialog modal the same way (i.e not in the project service, and without jQuery to fetch input contents).

FYI, we're probably not backporting anything to the old release (but it's fine to leave what you have there).

kcolford:
Alright, this one is exactly as you asked for (commit dialog model in frontend.js and no jQuery).  All the UI events are hooked up except the keyboard events, that might come in a later patch.

e45lee:
Looks fine to me.

PR #148, opened 2015-02-03
[frontend] Added alternative download dialog.

kcolford:
This is an alternative means by which someone can download their
assignment code.  It provides a dialog that gives them an SSH url that
can be used by either rsync or git.  The option to bring that up is
inside the dropdown menu along side the "download assignment" option.

### DIFF
--- a/src/collects/seashell/backend/dispatch.rkt
+++ b/src/collects/seashell/backend/dispatch.rkt
@@ -513,6 +513,13 @@
        `#hash((id . ,id)
               (success . #t)
               (result . #t))]
+      [(hash-table
+        ('id id)
+        ('type "getPath")
+        ('project project))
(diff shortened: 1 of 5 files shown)


e45lee:
I'm not inclined to merge this.  There are a few reasons for this:

1.  It'll confuse students.
2.  The layout of everything under ~/.seashell is not stable, and we don't intend on making
     it so.  For example, we may not be using git to store snapshots in a future release.
3.  These are really only 1-way (download) syncs, but there's nothing stopping an unsuspecting
     student from writing into ~/.seashell and getting the system into an inconsistent state.
     We'd rather this not happen.
4.  It's not that hard to launch the Seashell instance oneself (via SSH) and make the appropriate
     API calls to download an assignment anyways.

PR #159, opened 2015-02-11
[backend] Reduced network load by only transmitting diffs on save.

kcolford:
The current network load was too high for saving a whole file every
second.  Since a it is unrealistic for a person to edit their entire
file in that time, we compute a unified diff of the changes they've
made and then send only that over the network.  On the server side, we
fork to start the `patch' program which adjusts the file so it is
correct.

We've copied the script from jsdifflib <http://snowtide.com/jsdifflib>
into the VCS to enable us to compute the diffs on the client machine.
We've also added functionality for it to output properly formatted
diffs.  The only new outside dependency is that the `patch' program
must be installed on the host machine.


m4burns:
It seems a bit wasteful to invoke the `patch' program every second for every open editor; you might hit other resource limits (not to mention the added dependency on a command-line utility). Have you considered https://code.google.com/p/google-diff-match-patch/ ? This library is performant, very simple to use, and available in JavaScript and C.

kcolford:
The problem with that library is that it requires Qt, which is a massive library to be bringing in.  Although you are right, invoking patch every second will be a problem.  Maybe I can adapt the source code for the GNU Diffutils and then since it's in C we can just directly add in bindings through racket's FFI.

m4burns:
There is a port to C++/STL. Thread: https://code.google.com/p/google-diff-match-patch/issues/detail?id=25 code: https://github.com/leutloff/diff-match-patch-cpp-stl

kcolford:
Oh, excellent.  That's perfect.

m4burns:
I think this is also a good starting point for concurrent editing.

e45lee:
We should probably design how concurrent/offline mode will look like before messing with SeashellFile and adding an API call that also touches files.

FYI, we save 2 seconds after your last change to the CodeMirror buffer.

Is there a way to do this without adding another FFI binding or depending on an external process?

kcolford:
If you don't want an FFI binding or an external process then the only options left are a kernel module (which the worst option to take), writing it out in racket, or using an existing FFI binding (libgit2 has bindings for applying patches in this way).

m4burns:
Writing a simple text diff/patch that uses edit distance isn't too much work. Might be better than taking on a new dependency or the fragility of FFI bindings.

kcolford:
Ok then, so I'll throw something like that together.

kcolford:
I just pushed the new patch that does what's been discussed.  Currently it's heavily reliant on line numbers because that lets the network load be as small as possible, but we can easily adjust the way it encodes the diff to add more lines of context.

e45lee:
We're probably fine without lines of context; I expect that we'll use some other mechanism that will prevent us from applying diffs on the wrong version of a file.

kcolford:
Ok, then it should work like that as is.

e45lee:
One thing to note: You should probably deal with the case that if self.old_data is null, then you send the full file contents.

Other than that, this should be fine.

kcolford:
That edge case has been delt with by setting self.old_data to the contents of the file each time that SeashellFile.prototype.read is called (note the edit there).

e45lee:
It is possible (and should remain possible) to call write() before read().

kcolford:
Ok, it's been fixed by setting self.old_data to something not null in the case that it is null or undefined.

e45lee:
It's still more efficient then to call writeFile if you're diffing against '\n' in this case...

PR #164, opened 2015-02-12
[frontend] Made most recent file open up when assignment is selected.

kcolford:
Fixes issues #156 and #79

PR #165, opened 2015-02-12
[backend] Open most recent file when opening assignment.

kcolford:
follows up on #164


e45lee:
Sorry to nitpick again, but I'd also like the most recent question/file functionality split in the backend.  Something like get/set-recent-question (takes project name, returns question), and get/set-recent file (takes project name and question, returns file name and part)

kcolford:
I since the backend makes no distinction between projects/questions, it just considers directories and files, I felt it ought to keep up with that mentality. ‎Thus, when we open a file it updates all the directories above it (in this case 2, but I've been considering abstracting it since it feels out of place like that). Then you only check the most recent sub-path for a particular directory, this maintains the simplicity of the backend only being aware of the file system while the frontend handles the structure of assignments/questions/sources. 
‎

kcolford:
This patch should work nicely.

kcolford:
Bump.

This pull request has been idle for a while, so I would like to get some feedback on whether it should be merged/inproved/closed.

src/collects/seashell/backend/files.rkt
```diff
@@ -223,3 +225,47 @@
...
+
+  (with-output-to-file recent (thunk (write r)))
+  (void))
+
+;; (get-recent-file directory)
+;; Reads the most recent file name for the specified project.
+;;
+;; Arguments:
+;;  directory - the directory to check the default for.
+(define/contract (get-recent-file directory)
```
e45lee: One final thing; get-recent-file should take in a project, subdirectory pair for consistency.

kcolford:
Updated, just waiting on travis.

PR #166, opened 2015-02-12
[frontend] Open most recent file when opening assignment.

kcolford:
follow up on #164

### DIFF
--- a/src/frontend/frontend/frontend.js
+++ b/src/frontend/frontend/frontend.js
@@ -552,7 +552,7 @@ angular.module('frontend-app', ['seashell-websocket', 'seashell-projects', 'jque
         self.close = function () {
             $state.go('list-projects');
         };
-      self.project.resentFile()
+      self.project.recentFile()
         .then(function (recent) {
           if (recent) {
             $state.go('edit-project.editor',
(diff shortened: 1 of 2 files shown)


e45lee:
So the WebSockets returns a jQuery.Deferred object which will resolve to a value once a response is received from the backend - it does not return a value which can be directly used -- so your code as it stands will not work correctly here.

We typically wrap these Deferred objects with Angular's $q library (using $q.when - there are plenty of examples in the codebase).

You can interact with these $q Deferred objects using .then and .catch (to handle success/failure, respectively).  Take a look at https://docs.angularjs.org/api/ng/service/$q

kcolford:
This next patch addresses that

kcolford:
Done. Also, the code builds.

kcolford:
Bump.

This pull request has been idle for a while, so I would like to get some feedback on whether it should be merged/inproved/closed.

e45lee:
I'll take a look at it hopefully sometime this week.

src/frontend/frontend/frontend.js
```diff
@@ -552,6 +552,14 @@ angular.module('frontend-app', ['seashell-websocket', 'seashell-projects', 'jque
         self.close = function () {
             $state.go('list-projects');
         };
+      self.project.resentFile()
```
kpalway: Should be recentFile()

kcolford: Fixed with recent patch.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           ‎

PR #185, opened 2015-02-19
Set self.loaded to true after opening a file.

kpalway:
Fixes the not-saving-after-pasting issue

### DIFF
--- a/src/frontend/frontend/frontend.js
+++ b/src/frontend/frontend/frontend.js
@@ -1033,7 +1033,7 @@ angular.module('frontend-app', ['seashell-websocket', 'seashell-projects', 'jque
         self.project.openFile(self.question, self.folder, self.file)
           .then(function(conts) {
             self.contents = conts;
-            self.ready = true;
+            self.ready = self.loaded = true;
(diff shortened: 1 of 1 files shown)


e45lee:
So the reason self.loaded is set in the first on('change') event is that
 - the first 'change' event fired is usually triggered by the initial file contents load into the editor.
 - we need to clear the change history, so pressing Ctrl-Z right after the editor's ready doesn't
   undo the last change (which was to load the file contents) and clear the editor!

The first change event is probably not firing because we're changing the contents from an empty string
to an empty string, which is apparently special cased in CM: https://github.com/codemirror/CodeMirror/blob/de6848600acb7b6f22fc1d7ea0d4094ca5faeced/lib/codemirror.js#L3645

So only set self.loaded = true in the load handler iff the contents are the empty string.

kpalway:
Even if I change it back to how it was, Ctrl-z still clears the whole editor right after opening a file, so I'm not clear on the benefit of setting self.loaded later

e45lee:
Did you pull fixes from master recently?

PR #191, opened 2015-02-23
Reduced compilation time of LLVM by only building for the host as the target.

kcolford:


kcolford:
Hmmm, travis says that it only improves it by 3 minutes... I think I'll close this as not useful.

Issue #190, opened 2015-02-21
"Esc" in Vim mode being overriden by exit fullscreen shortcut


Gibstick:
Esc is set to exit full-screen, but this breaks Esc in Vim mode. See https://piazza.com/class/i3g6q2yjffi6td?cid=1742

Relevant lines might be https://github.com/cs136/seashell/blob/0079494ae5b17a8fe1f35b3dedc228d084f05eb5/src/frontend/frontend/frontend.js#L894-896

kcolford:
Hmmm, it appears that Vim mode breaks one of seashell's features... Another way in which Emacs is the better editor...

That aside, we could easily patch it by disabling seashell's Esc binding (at least while Vim mode is enabled), the codemirror documentation http://codemirror.net/doc/manual.html#addKeyMap shows that this later binding is in fact overshadowing the Vim mode one.

thadeaus:
Enter-ctrl toggles full screen. We can get rid of the lines that bind Esc to exit fullscreen that you highlighted. I didnt consider vim mode. Thanks for pointing this out.

kcolford:
I think this issue can be closed now

PR #192, opened 2015-02-24
[frontend] Prevent Esc from being overridden in Vim mode.

kcolford:
Fixes issue #190


e45lee:
This can leave the UI in a stuck state when in fullscreen.  Pick another key to bind here.

kcolford:
No it would not, because as was discussed in the thread on the issue, there is still the Ctrl-Enter binding that toggles the state of the full screen mode.

e45lee:
Alright, sure.

PR #195, opened 2015-02-24
Load Marmoset results on question open, if available

kpalway:
Had to change getting the results after submitting as well to deal with concurrency problems (making sure the project list loads before we get results). Haven't tested if submitting and getting results after still works the same, since one of the live environments is required for that.


src/frontend/frontend/frontend.js
```diff
@@ -650,6 +653,35 @@ angular.module('frontend-app', ['seashell-websocket', 'seashell-projects', 'jque
           self.common_files = result.common;
           self.question_files = result.question;
           self.test_files = result.tests;
+
+          self.project.currentMarmosetProject(self.question).then(function(target) {
+          if(target) {
+            marmoset.results(target).then(function(result) {
```
e45lee: Can we pull most of this (shared) code out into its own function, so we only have one copy of it in the codebase?

e45lee:
Looks fine, and I'll merge so you can test on the Linux environment, but remember to pull out the common code that deals with results into its own function.

PR #207, opened 2015-03-06
Allow for Manual Resize

thadeaus:
Changes the elements using a new activate Resize function. We can move where the button is but this seemed the most appropriate. 

One more thing that was commented on was that I could add it so we can make the console larger when we resize, but this can cause some weird issues. If you would like that implemented then let me know. For now it still all just remains on one screen-height.

### DIFF
--- a/src/frontend/css/common.css
+++ b/src/frontend/css/common.css
@@ -342,8 +342,13 @@ div.tooltip .tooltip-inner {
   text-align:right;
 }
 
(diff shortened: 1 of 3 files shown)


thadeaus:
As seen by the latest commit, the input bar is now resizing correctly. I gave it an initial width of 100% in the template file.

e45lee:
LGTM

PR #219, opened 2015-03-20
Made RUN/TEST context-sensitive

b2coutts:


e45lee:
Will merge for now, but in the future, avoid using $scope.$parent

PR #224, opened 2015-03-24
Saves Resize Option and Saves Previous Line Number

thadeaus:


e45lee:
You may as well persist this using the same infrastructure that also saves most recently used information.
This is a larger change, so I'll merge this for now, but we should talk.

PR #286, opened 2015-05-05
Bug fix: Safari file upload

kpalway:


e45lee:
There's a bug in the regular expression that parses data URL's.  I'll commit a more permanent fix soon.

PR #289, opened 2015-05-06
Update to clang 3.6

e45lee:
Update to clang 3.6, as well as sync the compiler code with the stuff we're going to compile down to JavaScript.


kpalway:
This would be a good time for us to write some tests for the compiler code.

e45lee:
Yeah, let's try to co-ordinate this.  I don't think I'll be able to find time until next week, though.

PR #309, opened 2015-05-14
Fixed Console/Editor Offset

thadeaus:
Do not remove the issue for this, as I still need to fix the resize issue when the errors pop up.


e45lee:
I'm not entirely sure this fixes it, as we had this code before, and it didn't always work.  Granted, the current code doesn't always work either.

Can you test on other machines to see if this fixes the issue?  I remember that before (and now, I think), the +1 (or lack of) would cause different behaviour on different machines; some would be aligned, some would be 1 pixel too tall, some would be 1 pixel short.

e45lee:
Anyways, let's give this a shot for now.

thadeaus:
Ok I will try it in all the different browsers in the next few days. Also, Is there an area in the code I can look at so the error messages don't immediately pop up? I think they are causing one of the issues.

e45lee:
OK, the resize thing was a jQuery issue.  You could probably just hide all the error boxes, and/or trigger a resize event with $timeout(..., 0) to fix your other issue.

PR #312, opened 2015-05-15
[frontend] Remove jQuery calls; hopefully fix off-by-one height issues.

e45lee:
This hopefully fixes the off by one height issues we've been having for a while now.  jQuery's height function didn't always set the height properly for us (it set it to +/- 1px sometimes).


kpalway:
Is there any reason to think that this was a problem with jQuery? ie. does this code work differently?

e45lee:
Well, other than the fact that this works for me, with no (+/-1) height adjustments, on one browser/OS configuration where the problem occurred before.

(YMMV, though.  I never saw this problem on the school computers, mind you -- but we needed to adjust the console's height by 1px for stuff to work there)

kpalway:
Alright, let's try it

PR #331, opened 2015-06-06
Multiple CSS Themes

thadeaus:
Console still does not get updated, e45lee was going to check that out later, I am just pushing this up so that it is visible that it is ready to be pulled (aside from the console not updating).

Tappy made almost all of the changes, I was just getting it working with our newest builds and changed the css on the bootstrap buttons.


e45lee:
OK, I will merge this from my end.

PR #358, opened 2015-08-17
Rename between questions/folders

kpalway:


src/frontend/frontend/modals.js
```diff
@@ -138,7 +138,7 @@ angular.module('frontend-app')
           templateUrl: "frontend/templates/rename-file-template.html",
           controller: ["$scope", "$state", "error-service",
             function($scope, $state, errors) {
-              $scope.rename_name = file;
+              $scope.rename_name = question + "/" + (folder=='question' ? "" : folder + "/") + file;
```
e45lee: I don't think this works for moving in/out of the common/ folder....

e45lee:
Looks mostly fine, except I don't think you handle the common/ folder case properly.  Can you take a look at this?

kpalway:
Right. Fixed that now. There is still some weirdness because of discrepancies with how we display the filesystem vs. how it actually is stored, ie. common folder path is common/file.c when the user sees q1/common/file.c.. for now we can probably just leave this up to the user.

PR #376, opened 2015-09-29
Redirect to CAS login when websocket fails to connect

b2coutts:


e45lee:
Note: roll back this pull request when we finally implement offline mode.

Issue #378, opened 2015-09-30
When uploading an empty file, an error occurs.


Gibstick:
match-define: no clause for #f

Gibstick:
Only happens on chrome, because it sends "data:"

PR #382, opened 2015-10-01
Fix empty file bug

b2coutts:
When the file data parsing regex fails to match, produces an empty string
instead of crashing


Gibstick:
Fixes #378

PR #401, opened 2015-11-06
Create and return checksums for files.

e45lee:


e45lee:
These changes will land as part of offline-mode, closing.

PR #415, opened 2015-11-09
remove asan and use valgrind

Gibstick:
Testing only


m4burns:
Nice.

Gibstick:
Re-doing the tests because of the pre-existing race condition that caused them to fail.

PR #417, opened 2015-11-10
Fixed .in files >> stdin race condition

b2coutts:
With large.in files, and programs that do not read the input and terminate
before write-bytes can finish writing the .in file to the program stdin,
seashell would try to write to the closed stdin port, causing an error.


Gibstick:
Also fixes the tests failing when we run the programs under Valgrind :D

PR #420, opened 2015-11-11
Make ASan detect stack use-after-return

Gibstick:


Gibstick:
This literally puts stack frames on the heap. https://github.com/google/sanitizers/wiki/AddressSanitizerUseAfterReturn

PR #429, opened 2015-12-17
Runnerguy

Gibstick:


src/collects/seashell/backend/project.rkt
```diff
@@ -316,13 +323,69 @@
...
+;; Arguments:
+;;  main-file - The .c file being compiled
+;;  file-dir  - The directory containing main-file
+;;
+;; Returns:
+;;  A list of the .h files included by a program, with the .h extension stripped
+;; Raises:
+;;  exn:project if an included header is not a .h file
+;; TODO: need to clean up the subprocess and ports?
+(define/contract (get-headers main-file file-dir)
```
e45lee: You are going to want to clean up/close the I/O ports you get back from subprocess here.  The GC will not automatically close OS-level ports.

src/tests/tests/compiler.rkt
```diff
@@ -29,16 +29,17 @@
...
-      (new-project "foo3")
-      (new-file "foo3" "2.c" #"#include <stdio.h>\nvoid a() {\nprintf(\"Hello.\");\n}\n" 'raw #f)
-      (new-file "foo3" "1.c" #"int main() {a(); return 0;}" 'raw #f)
-      (define-values (success hsh) (compile-and-run-project "foo3" "1.c" '()))
-      (check-true success)
-      (define pid (hash-ref hsh 'pid))
-      (sync (program-wait-evt pid))
-      (check string=? (port->string (program-stdout pid)) "Hello.")
-      (delete-project "foo3"))
+     ;; TODO: Fix this test, because compile-and-run-project has changed
```
e45lee: Fix the test first.

src/frontend/frontend/file.js
```diff
@@ -384,7 +380,7 @@ angular.module('frontend-app')
         self.runFile = function() {runWhenSaved(function () {
           self.killProgram().then(function() {
             self.console.clear();
-            self.project.run(self.question, "question", self.runnerFile, self.contents, false)
+            self.project.run(self.question, "question", self.file, self.contents, false)
```
e45lee: This is confusing, as self.file is not really used (outside of the first run).  We should remove this parameter entirely.

src/frontend/frontend/file.js
```diff
@@ -406,7 +402,7 @@ angular.module('frontend-app')
         self.testFile = function() {runWhenSaved(function () {
           self.killProgram().then(function() {
             self.console.clear();
-            self.project.run(self.question, "question", self.runnerFile, self.contents, true)
+            self.project.run(self.question, "question", self.file, self.contents, true)
```
e45lee: (See above)

src/collects/seashell/backend/project.rkt
```diff
@@ -440,6 +488,22 @@
     [else
       (values #f `#hash((messages . ,messages) (status . "compile-failed")))]))
 
+
+;; (compile-and-run-project/use-runner name tests)
+;; is a wrapper around compile-and-run-project, supplying the file
+;; from the project settings file.
+;; Note: the file arg is used only to determine the question
+(define/contract (compile-and-run-project/use-runner name file tests)
```
e45lee: I would prefer if we got the "question" directory here instead of breaking up the file path to find the question directory.

src/collects/seashell/backend/project.rkt
```diff
@@ -440,6 +488,22 @@
...
+  (-> project-name? path-string? (listof path-string?)
+      (values boolean?
+              hash?))
+  (define question (car (string-split file "/")))
+  (define file-to-run (get-file-to-run name question))
+  (if (string=? file-to-run "")
```
e45lee: Toss an error here instead, and have the frontend ensure that the runner file is set properly?

src/collects/seashell/backend/project.rkt
e45lee: Rename file -> file-for-location, so it's more descriptive.

src/frontend/frontend/file.js
e45lee: Can we omit file entirely, as it's kind of confusing to see it here when we don't actually depend on it (we really only depend on the question now)

src/frontend/frontend/file.js
e45lee: See above.

src/collects/seashell/backend/project.rkt
```diff
@@ -316,13 +323,69 @@
...
+;;
+;; Arguments:
+;;  headers - The list of included local .h files, i.e., produced by get-headers
+;;
+;; Returns:
+;;  A list of the .c files and a list of the .o files to be compiled/linked with a program.
+;; Raises:
+;;  exn:project if an element of headers is not a .h file, if a .h file has no
+;;  corresponding .o or .c file, or if a .h file has both a .c and .o file.
+(define/contract (get-co-files headers)
```
e45lee: Can you guys write a test for this?

e45lee:
Hey, I've left a few comments on the code so far.  It looks mostly fine, though there are a few confusing bits (which should be addressed), and there are missing tests (which should be written, as we'd rather not find out that everything's broken in January).

Gibstick:
Sure thing. I'll fix the sloppy code after the final.

e45lee:
Looks fine.

Issue #437, opened 2015-12-23
Allow runtime-files-path to be configurable


e45lee:
The compiler stores binaries in (runtime-files-path).  This should be configurable so the CLI tool doesn't need ~/.seashell/runtime-files to exist.

yc2lee:
Do you have a preference for where the user would configure this? (some config file setting or option command line argument or something else...)

e45lee:
We store settings in https://github.com/cs136/seashell/blob/master/src/collects/seashell-config.rkt.in

e45lee:
Nick fixed this as of #438.

PR #438, opened 2015-12-23
put runtime-files in temporary folder when using using seashell-cli

yc2lee:


e45lee:
Don't we want the runtime-files directory in /tmp in the CLI tool?

Gibstick:
@e45lee I thought this was preferable, so I told him to do it this way. But /tmp is fine too.

e45lee:
I have a slight preference to /tmp here, as avoiding frequent network writes is preferable (~cs136t is mounted over NFS).  It would be nice if this was configurable, see #437.

e45lee:
Looks good, I'll merge and fix the spelling mistake, thanks.

Issue #404, opened 2015-11-06
Ctrl+u is bound to undo as well as test


Gibstick:

YLiLarry:
Rebound the hotkey to Control+Shift+R

YLiLarry:
Need a suggestion of this hotkey!

YLiLarry:
Using Control+E
Pull request #439

PR #439, opened 2015-12-24
Fix Issue #404: Control+U is bound to undo as well as test

YLiLarry:
Issue #404
Rebound test hotkey to Control+E

### DIFF
--- a/src/frontend/frontend/file.js
+++ b/src/frontend/frontend/file.js
@@ -34,7 +34,7 @@ angular.module('frontend-app')
         self.project = openProject;
         self.question = openQuestion;
         self.folder = openFolder;
-        self.file = openFile; 
+        self.file = openFile;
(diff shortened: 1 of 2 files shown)


Gibstick:
Ctrl+Shift+R is already bound to regex search in CodeMirror

YLiLarry:
Are you sure? I tried that combo on my mac and nothing happened, so I chose it. 
Anyways, any idea what could be a good combo?

e45lee:
That might be AppleKey-Shift-R on the Mac.  Try Ctrl-E (evaluate), if that's not already bound?

e45lee:
One quick note: we land merges in master.  Please have your PR target that branch, thanks.

Gibstick:
Don't worry too much about this, Larry. Bryan gets confused about the
branches every week.

PR #441, opened 2015-12-25
Prevent creating undeleteable projects

### LINKED ISSUE #364: Disallow creating undeletable assignments.

Gibstick:
Fixes #364


Gibstick:
Should deleteability be a per-project setting now? I think it makes sense to move this to the project settings file.

e45lee:
Probably, and then we don't have to merge this code in either.

PR #451, opened 2016-01-01
[frontend] hotfix regression that broke tests

Gibstick:
When compileAndRunProject was changed, the branch of the code that
dealt with running with tests  wasn't changed accordingly.

Hotfix version of #449


e45lee:
#449 already merged.

PR #455, opened 2016-01-04
Regex for placing "Section" projects (examples from notes) in their own section

Gibstick:
Might have some other stuff to do, like renaming "Lectures".


Gibstick:
Oops

Issue #456, opened 2016-01-05
Racket common files.


e45lee:
Hi everyone,

The Seashell team met yesterday and discussed all of the input we've
received over the past couple weeks. Here's an explanation of the
tentative decisions we made so there are no surprises:

"/common" Folder: There is currently a discrepancy as to how files in
the /common folder can be accessed in Racket vs. in C. Our solution
will be to "flatten" the structure so all files in the /common folder
are treated as if they are in each individual question folder (so we
will have to copy these when we compile/run). This is consistent with
what currently happens when the student submits to Marmoset. We will
have to enforce that no question folder can contain a file with the
same name as a file in /common.

Refreshing Assignment Skeletons: When a project is opened, Seashell
will search for any files from the skeleton that are missing, and
replace any that it does not find with the fresh file from the
skeleton (perhaps a button will be added to do this without re-opening
the project). This way, if a student wants to get the fresh file, they
can rename or delete their existing file and get the fresh one. If the
course staff needs to change a skeleton file, there are a couple of
options: Change the file in the skeleton and request that all students
rename their in-progress file and refresh; or add a new file to the
skeleton which will be loaded the next time the student opens the
project. Either way the work of merging these changes is done by the
student.

If you have any questions or concerns about these points, let me know.

Thanks,
Kaleb

YLiLarry:
Ok I need meet up with you guys in campus. Where can I find you?

yc2lee:
Hey @YLiLarry I'm supposed to help with this issue. We can meet in my office or somewhere else if you want

YLiLarry:
I can go to your office. Do you have time after 3?

yc2lee:
I'll be in my office at 3 but I might not have time until 4 today

YLiLarry:
Ok I'll see you at 4. Where is your office?

yc2lee:
Hey Larry and I cannot find a command-line flag to /usr/bin/racket that would allow students to write something like (require "file.rkt") instead of (require "../common/file.rkt").

So, I am going to do exactly what Kaleb says. When students run a racket file, I'll create a temp directory, copy all the common files and question files there, and run /usr/bin/racket from there.

Please let me know if this will be as issue.

e45lee:
I imagine that should work, though it might lead to some undefined behavior
with regards to files in the question directories if the student's code
writes to files there.

You'll want to make sure the current-directory parameter is set up to point
to the question folder not the temp folder.

yc2lee:
[...]

PR #461, opened 2016-01-13
fix semicolons

YLiLarry:


YLiLarry:
frontend fix for issue #456

Issue #430, opened 2015-12-21
Add ability to reset file to original skeleton.


cs136:

e45lee:
https://github.com/cs136/seashell/compare/master...e45lee:revert-templates?expand=1

PR #470, opened 2016-01-17
New feature: reset to skeleton #430

YLiLarry:
ticket #430


YLiLarry:
This feature is done.

e45lee:
There are a few issues with this PR:
 1) There's already common infrastructure with dealing with template ZIP files - see backend/template.rkt.
 2) Why are you creating a new API endpoint for reading the list of files from a ZIP file?  There's
     already code written for that on the server.
 3) Where do we fetch the template contents?  I don't see any code written for this....

YLiLarry:
1) What branch are you at? I use master and I don't see that file.
2) same
3) What do you mean by template contents?

e45lee:
Eh, you'll want to fetch from e45lee/revert-templates.  I'm wondering where exactly do we populate the files with their contents from the skeleton.  The code I see creates empty files, unless I'm missing something here.

YLiLarry:
Well, frontend sends requests to the live server (cross-domain request) to get the file contents form zip url, then sends newFile request to the backend. To solve the cross-domain issue, I wanted to have a cgi on the live server, but since I don't have it I wrote that "readFileFromZipURL" in the backend.

YLiLarry:
Does restore-file-from-template create a file if it doesn't exists in the local project, and overwrites the existing file if already exists? guess I can merge my changes with yours.

YLiLarry:
Frontends holds a list of skeleton projects, and the projects skeleton file paths.
It checks that if the file the user deletes is in the skeleton. 
If so it sends "delete" then "readFileFromZipURL", then sends "newFile" request, otherwise it sends only "delete".
If restore-file-from-template both downloads and overwrites the file, I can send "restoreFileFromTemplate", then I should receive the file contents, so that I can create it in the frontend.

e45lee:
Looks good.

PR #472, opened 2016-01-19
flatten racket question and common files into one temp folder

yc2lee:
Fixing issue #456 

I wrote this code to flatten the files in question folder and common folder into one big temporary folder. I assume that common folder won't have folders inside it.

Now if students get an error, their error will look a bit more ugly because the file path has seashell-racket-temp-blahblahnumbers in it. Not sure how big of an issue this is

### DIFF
--- a/src/collects/seashell/backend/project.rkt
+++ b/src/collects/seashell/backend/project.rkt
@@ -479,32 +479,49 @@
                     diagnostics)))))
       (values result parsed-messages output-path))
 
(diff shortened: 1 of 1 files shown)


Gibstick:
Looks good, especially since it passes https://github.com/cs136/seashell/blob/master/src/tests/tests/project.rkt#L75

e45lee:
Incidentally, we don't have any Racket project tests.  We should probably write some.

PR #474, opened 2016-01-19
Fixed one off error when underlining errors

Tiedye:
Error is consistent across the most recent versions of Chrome, Firefox, and Edge

Image of Bug


Gibstick:
I can't find a trivial test case that gives off-by-one errors for me. Can you paste a simple test case? If it's your assignment, you can email it to seashell@cs.uwaterloo.ca.

Gibstick:
Looks good to me.

PR #483, opened 2016-01-31
80 char alert

YLiLarry:
Everyone happy with this color?

qq20160203-0
qq20160203-1

### DIFF
--- a/src/frontend/css/common.css
+++ b/src/frontend/css/common.css
@@ -414,7 +414,12 @@ input::-ms-clear {
 }
 
 .col-too-long {
-  color: rgb(240, 0, 0);
+  color: #DB2D20; 
+}
+
+.cm-line-too-long {
+  background-color: #E91E63 !important; 
+  color: white !important;
 }
 
 /* navbar overrides */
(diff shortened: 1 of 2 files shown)


Gibstick:
There are performance issues with this once you get past a small number of lines. They might have huge test files, and we don't want to break those. Also, on slower computers typing becomes janky. Can you look into doing this by only looking at the current line and dealing with paste events?

YLiLarry:
How about a periodical check per 5 seconds?

Gibstick:
I don't think five seconds will help. It will just be a pause every five
seconds.

If it's too difficult to do what I suggested above, then go for only
highlighting the current line.

YLiLarry:
Wait I thought you meant only highlight the current line. 
You meant only check 80 chars in the current line while keeping other highlighted lines unchanged, unless there's a paste event? Sure that's not hard.

Gibstick:
Yes, that's what I mean. Only check what you have to. And I guess you have
to check the entire file when it is first loaded.

YLiLarry:
Good advice. I'm on it.

PR #487, opened 2016-02-05
Common to common includes

Gibstick:
Caveats:
- common to uncommon does not work (it never did)
- uncommon to common to uncommon does not work (corollary of above)


e45lee:
I imagine that adding "-I <question_dir>" will solve the common -> uncommon problem, but let's merge this for now.

PR #489, opened 2016-02-05
Merge master into stable for 2.5.5

Gibstick:


Gibstick:
The test failure isn't actually a test failure. I merged too quickly for Travis.

PR #498, opened 2016-02-08
Project whitelist for users

YLiLarry:
Projects whitelist is in ~cs136/assignment_skeletons/project_whitelist.json
User whitelist is ~cs136/assignment_skeletons/user_whitelist.json

I also added their cgis in ~cs136/cgi-bin/. Not sure how to execute them tho.

I still have cross domain request issue when I try to test it on ugister, so alternatively I tested by imitating the server response. It should work on the live server.

### DIFF
--- a/src/frontend/js/project-service.js
+++ b/src/frontend/js/project-service.js
@@ -34,8 +34,8 @@ angular.module('seashell-projects', ['seashell-websocket', 'marmoset-bindings'])
       // TODO: update with real template path.
       var PROJ_ZIP_URL_TEMPLATE = SKEL_ROOT_URL + "%s-seashell.zip";
       var PROJ_FILE_LIST_URL_TEMPLATE = CGI_URL + "skeleton_file_list.rkt?template=%s";
-      var PROJ_WHITE_LIST_URL = SKEL_ROOT_URL + "project_whitelist.json";
-      var USER_WHITE_LIST_URL = SKEL_ROOT_URL + "user_whitelist.json";
+      var PROJ_WHITE_LIST_URL = CGI_URL + "project_whitelist.cgi";
+      var USER_WHITE_LIST_URL = CGI_URL + "user_whitelist.cgi";
      
       var SeashellProject = (function () { 
         /**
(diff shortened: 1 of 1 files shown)


e45lee:
Try wrapping the JSON files in a cgi-script that outputs "Access-Control-Allow-Origin: *".  You will have to edit .htaccess.

YLiLarry:
Hi Edward, don't I have to restart the https://www.student.cs.uwaterloo.ca/ apache server to make that effective?

src/frontend/js/project-service.js
```diff
@@ -34,6 +34,8 @@ angular.module('seashell-projects', ['seashell-websocket', 'marmoset-bindings'])
       // TODO: update with real template path.
       var PROJ_ZIP_URL_TEMPLATE = SKEL_ROOT_URL + "%s-seashell.zip";
       var PROJ_FILE_LIST_URL_TEMPLATE = CGI_URL + "skeleton_file_list.rkt?template=%s";
+      var PROJ_WHITE_LIST_URL = SKEL_ROOT_URL + "project_whitelist.json";
```
e45lee: Shouldn't this be a CGI script?

e45lee:
No; also, you should run the HTTP request against the CGI script.

YLiLarry:
Yeah I will change them to cgi scripts.
Can you take a look at why https://www.student.cs.uwaterloo.ca/~cs136/cgi-bin/user_whitelist.cgi is displayed as a plain text?
I'm not sure how the server is configured. yet.

YLiLarry:
:3 
Looked into `cat /etc/apache2/sites-available/default` but nothing useful found.

Gibstick:
I fixed the .htaccess in ~cs136/public_html/cgi-bin/ and it executes now.

YLiLarry:
How did you do that? What did you change... :3

Gibstick:
You have to specifically whitelist every cgi script in there. See the very bottom of the file. And make sure you look in ~/www/cgi-bin/.htaccess and not ~/www/.htaccess

YLiLarry:
Works like a magic. Thanks Charlie.
This feature is finished.

PR #506, opened 2016-03-01
Typed Racket Conversion

e45lee:
This PR partially converts the backend to Typed Racket.  More to follow.


e45lee:
This commit partially converts the backend to Typed Racket.  What's still left as untyped racket is the code in seashell/backend.  Some work was also done with regards to cleaning up the subprocess/custodian situation in seashell/tunnel and seashell/backend/runner.

PR #508, opened 2016-03-08
Make common headers included with <...> tags work in Marmoset

yc2lee:
This commit makes the student's zip file that's uploaded to Marmoset have the common/ folder. The CS136 ISAs will need to copy the common/ folder to the /tmp/$$.dir folder in their testing script for it to actually work.

### DIFF
--- a/src/collects/seashell/backend/project.rkt
+++ b/src/collects/seashell/backend/project.rkt
@@ -650,6 +650,7 @@
             
             (copy-from! question-dir)
             (when (directory-exists? common-dir)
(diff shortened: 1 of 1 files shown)


e45lee:
Do we still want to merge common into the question folder?

yc2lee:
Not sure. I left it there in case CS136's Marmoset testing scripts use those files for some reason and I'm worried some students will do weird things wherein if those files are gone, they'll no longer pass tests.

Should I delete https://github.com/cs136/seashell/blob/master/src/collects/seashell/backend/project.rkt#L653 ?

e45lee:
I think so; this would only affect retests.

PR #512, opened 2016-03-28
Asan error parser (now integrated with frontend)

YLiLarry:
#517 affects this.
Besides that the asan error message is now parsed/reformatted by the backend.
The way I achieved this is to put asan log into a ".asan.{pid}" file in {project}/{question} directory.
The file is read and sent along with the "type: done" transmission from the backend.
I haven't removed the old javascript parse code. Maybe this is what I need to do next, or I can just leave it.
Also the new asan format needs to be prettified.


e45lee:
You'll want to look here for the flags to extract the ASAN error out of
stderr: https://github.com/google/sanitizers/wiki/SanitizerCommonFlags

For now, yes, I'd like the parser moved to the backend.  It would also be
useful if the parser could also parse the error information out of ASAN
(like buffer-overflow @<addr>, X bytes into frame....)

On Mon, Mar 28, 2016 at 1:45 PM, Yu Li <notifications@github.com> wrote:

> Sorry about this late push.
> I've a bit confused what I'm suppose to do.
> I couldn't find a flag that separates out solely the AsanSanitizer error
> from the linker's output.
> I can only put entire linker error to a file somewhere. I'll dig into this
> further more.
> Also Edward you should take a look if this parser is something you want.
> My current understanding is that you just want to move the parser code from
> the frontend js to backend typed racket?
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/cs136/seashell/pull/512
> Commit Summary
>
>    - Changed the format of asan frame output from single quote to json
>    double quote.
>    - merge master
>    - basic asan error parser
>
> File Changes
>
>    - *A* src/collects/seashell/backend/asan-error-message.rkt
>    <https://github.com/cs136/seashell/pull/512/files#diff-0> (97)
>    - *A* src/collects/seashell/backend/asan-error-parse.rkt
>    <https://github.com/cs136/seashell/pull/512/files#diff-1> (13)
>    - *M* src/collects/seashell/backend/runner.rkt
>    <https://github.com/cs136/seashell/pull/512/files#diff-2> (6)
>    - *M* src/frontend/frontend/console.js
>    <https://github.com/cs136/seashell/pull/512/files#diff-3> (1)
>
> Patch Links:
>
>    - https://github.com/cs136/seashell/pull/512.patch
>    - https://github.com/cs136/seashell/pull/512.diff
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/cs136/seashell/pull/512>
>



-- 
Edward Lee

e45lee:
Instead of parsing the error message in dispatch.rkt, can you parse it in runner.rkt and pass the result to dispatch.rkt?

src/collects/seashell/backend/dispatch.rkt
```diff
@@ -108,10 +112,21 @@
             (send-message connection `#hash((id . -4) (success . #t)
                                             (result . #hash((pid . ,pid) (test_name . ,test-name) (result . ,test-res)))))]
            [(list pid test-name "error" exit-code stderr)
+            ;; check for asan error message in "{project}/{question}/.asan.{pid}"
```
e45lee: Invoke the parser in runner.rkt and pass the result back from there.

src/collects/seashell/backend/dispatch.rkt
```diff
@@ -203,13 +218,22 @@
                     (program-kill pid)
                     (program-destroy-handle pid)]
                    [(? (lambda (evt) (eq? evt wait-evt)))
+                    ;; check for asan error message in "{project}/{question}/.asan.{pid}"
```
e45lee: Invoke the parser in runner.rkt and pass the result back from there.

src/collects/seashell/backend/runner.rkt
```diff
@@ -312,8 +312,11 @@
                   ;; Behaviour is inconsistent if we just exec directly.
                   ;; This seems to work.  (why: who knows?)
                   (putenv "ASAN_OPTIONS"
-                          "allocator_may_return_null=1:detect_leaks=1:stack_trace_format=\"{'frame': %n, 'module': '%m', 'offset': '%o', 'function': '%f', 'function_offset': '%q', 'file': '%s', 'line': %l, 'column': %c}\"
-                            detect_stack_use_after_return=1")
+                          "allocator_may_return_null=1:
+                           detect_leaks=1:
+                           log_path='.asan':
```
e45lee: Log to /tmp preferably.

src/collects/seashell/backend/asan-error-parse.rkt
```diff
@@ -0,0 +1,113 @@
...
+    (filter-map (lambda ([x : String])
+                  (match (string-trim x)
+                    ["" #f]
+                    [x x]))
+                (string-split contents "\n")))
+  (when (< (length lines) 2)
+    (error 'ParseError "Input message is too short.\n Here is an example of an expected input:\n 
+           =================================================================
+           ==33482==ERROR: LeakSanitizer: detected memory leaks
+           Direct leak of 144 byte(s) in 9 object(s) allocated from:
```
e45lee: Can you also parse the extra information (i.e the stack frame layout, location of fault) that ASAN produces?

src/frontend/js/socket/websocket_client.js
```diff
@@ -458,10 +458,11 @@ SeashellWebsocket.prototype.marmosetSubmit = function(project, assn, subdir, def
     deferred);
 };
 
-SeashellWebsocket.prototype.startIO = function(project, pid, deferred) {
+SeashellWebsocket.prototype.startIO = function(project, question, pid, deferred) {
   return this.sendMessage({
     type : "startIO",
     project : project,
+    question : question,
```
e45lee: I don't think you need to modify the API here.

e45lee:
Also, can you add tests for the ASAN parser?

src/frontend/js/socket/websocket_client.js
YLiLarry: You're right. I only needed to know the question because I logged to the question folder. If I log to /tmp I don't need to modify this. Thanks Edward!

src/collects/seashell/backend/dispatch.rkt
YLiLarry: Do you want the error message still in stderr if its invoked from runner.rkt? In dispatch.rkt I added a new filed for sanitizerMsg in the "type: done" api because I thought we could potentially separate it out from user fprintf(stderr ..) and we could do some highlight.

e45lee: Pass the error message in the list.  (Note: you'll have to use serialize,
look at the compiler code to see how that works)

src/collects/seashell/backend/runner.rkt
YLiLarry: Ok. I'll log to "/tmp/asan.{pid}".

src/collects/seashell/backend/dispatch.rkt
YLiLarry: Alright I'll change that. Just let you know that the asan message still needs to be sent back to the frontend through the "type: done" api if user runs their code without tests, and the compilation is success. (in which case the only options are to send back through "type: done", or through the last "type: stderr")

e45lee: That's fine.

YLiLarry:
Fixing

Issue #471, opened 2016-01-18
Allow setting a file in common as runner file


Gibstick:

yc2lee:
I created this on Kanban flow

yc2lee:
@e45lee Hey Ed I noticed that you cannot create a file named X in the question folder if X is already in common/. BUT if X is in the question folder, you can still create X in the common folder. Is this intentional?

Basically I'm asking if the same filename can appear in a question folder and in the common folder.

YLiLarry:
We can't prevent instructors from updating the common folder because the student has the same filename though.

Larry

YLiLarry:
It's more like I didn't care about that case when I wrote the script. If you want to prevent in two directions, we can always find a work around.

PR #521, opened 2016-04-21
Make common files runnable.

yc2lee:
#471

Hey @e45lee I made this commit to make common files runnable. I had to change a lot to make the code differentiate between files in common and question folder.

The project settings is ~/.seashell/projects/*/project-settings.txt has "common/filename.c" or "qX/filename.c" instead of just "filename.c"

I added optional parameters to functions for specifying the question because when I run from common/ folder, it doesn't have information on which question is actually being run. I need the question name for tests which are in the question folder and to -I include the right directory so that `#include "..."` will work.

A problem is that if students have the same racket filename in common/ and qX/, the one in qX/ will always be run. I'm not sure how to handle this because in a previous issue, we flatten the files. Now that common files can be run this might be a problem.


e45lee:
That's annoying (the Racket runner issue).  It's probably still worthwhile then to disallow having files with the same name in common/ and in the question folder.  Do you want to do that also in this pull request?

yc2lee:
Um I don't think I'll have time with Spring term starting. I probably won't get to it until after the first week of classes. Can @YLiLarry help? He wrote the code to prevent students from creating Y.c in qX/ if Y.c is in common/. I imagine doing the vice versa shouldn't be too hard?

YLiLarry:
Ok. I'll do this.

e45lee:
Hey I'm going to have to land #525 in front of this, which will result in some merge conflicts for this PR.  Can you rebase you work on #525?  You'll find that there's functionality for dealing with manually setting the test directory, so that should be helpful for you guys.

yc2lee:
Hey @YLiLarry where can I find the code for making sure that students cannot have the same filename in common/ and question folder?

YLiLarry:
This is urgent to fix? I'm bumping my head against the cli bug atm. It's in the frontend/frontend/modals.js, or you can search "Please avoid using a filename that exists in the "common/" directory." in the frontend code.

yc2lee:
Hey sorry for the delay. I merged the changes and I also disallowed creating file named X if X is already in question or common folder.

PR #542, opened 2016-05-13
Asan error cleanup 

YLiLarry:
Added an asan error parser to the backend.
Changed the format of TEST output when the tests fail.
Rewrite ASAN message to something less scary for students.

### DIFF
--- a/src/collects/seashell/backend/asan-error-parse.rkt
+++ b/src/collects/seashell/backend/asan-error-parse.rkt
@@ -1,0 +1,115 @@
+#lang typed/racket
+
+;; Use the function "asan-error-parse" to reads the asan error message
+;; and procude a parsed structure
+(provide asan-rewrite)
+
+(struct StackFrame ([frame : Integer]
+                    [file : String]
+                    [ln : Integer]
+                    [func : String])
+  #:transparent)
+(define-type ErrorComponent (U String StackFrame))
+(define-type ASANError (Listof ErrorComponent))
+
+(: asan-rewrite (Bytes -> Bytes))
+(define (asan-rewrite x)
+  (string->bytes/utf-8 
+    (ASANError->string 
+      (asan-error-parse (bytes->string/utf-8 x)))))
+
+;; rewrite ASANError in a friendly way
+(: ASANError->string (ASANError -> String))
+(define (ASANError->string x)
+  (string-join 
+    (map (lambda ([x : ErrorComponent])
+           (match x
+             [(? string?) x]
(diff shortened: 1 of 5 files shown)


src/collects/seashell/backend/asan-error-parse.rkt
```diff
@@ -0,0 +1,129 @@
...
+                          (asan-error-parse (bytes->string/utf-8 x)))))]))
+
+;; rewrite ASANError in a friendly way
+(: ASANError->string (ASANError -> String))
+(define (ASANError->string x)
+  (string-join 
+    (filter-map 
+      (lambda ([x : ErrorComponent])
+        (match x
+          [(pregexp #px"^=+\\d+=+ERROR: AddressSanitizer: SEGV") "Error: Segmentation fault"]
```
e45lee: Can you grab more detailed error information?  Ideally you could replicate the functionality from https://github.com/cs136/seashell/blob/master/src/frontend/frontend/console.js#L65.

src/collects/seashell/backend/asan-error-parse.rkt
```diff
@@ -0,0 +1,129 @@
...
+;; wrapper function
+(: asan-rewrite (Bytes -> Bytes))
+(define (asan-rewrite x)
+  (cond [(zero? (bytes-length x)) x]
+        [else (string->bytes/utf-8
+                (format "\n~a\n" 
+                        (ASANError->string 
+                          (asan-error-parse (bytes->string/utf-8 x)))))]))
+
+;; rewrite ASANError in a friendly way
```
e45lee: I'd like this to return JSON and have the frontend display the error message (so we can get line number information to indicate error locations)

src/frontend/frontend/console.js
```diff
@@ -213,47 +214,53 @@ angular.module('frontend-app')
     socket.register_callback("io", function(io) {
       if(io.type == "stdout") {
         ind = io.message.indexOf("\n");
-        if(ind > -1) {
+        if (ind > -1) {
```
e45lee: You should probably remove all the parse_asan code now that we don't need/use it here.

src/tests/tests/asan-error-parse.rkt
```diff
@@ -0,0 +1,6 @@
+#lang racket
+
+(require "../../collects/seashell/backend/asan-error-parse.rkt")
+
+(display (asan-rewrite (port->bytes)))
```
e45lee: Please write tests and integrate them into the existing testing framework we have.

PR #557, opened 2016-07-06
Max save snapshots

mbardakov:
Saving snapshot functionality (with "smart" snapshots)

### DIFF
--- a/src/collects/seashell/backend/backup.rkt
+++ b/src/collects/seashell/backend/backup.rkt
@@ -7,13 +7,15 @@
          racket/file
          racket/path
          racket/generator
-         racket/string)
+         racket/string
+         racket/set)
 
 (provide list-backups
          write-backup
          restore-from-backup
          write-backup-if-changed
-         get-backup-path)
+         get-backup-path
+         clean-backups)
 
 ;; (list-backups project file)
 ;; Lists all backups corresponding to a project+file. Sorted with the most recent first.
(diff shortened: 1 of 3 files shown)


e45lee:
Can you pull the backup stuff into another Racket module?  I want to minimize the number of changes to files.rkt if possible, as the offline mode code also touches files.rkt too.

e45lee:
Hey, can we do something here with getting rid of old backups?  Since we write a backup everytime a change is made the backups will get cluttered quickly.  Something like keeping at most A backups made in the last minute, B backups made in the last hour, C backups made in the last week would be nice...

src/collects/seashell/backend/backup.rkt
```diff
@@ -0,0 +1,111 @@
+#lang racket/base
+
+(require seashell/backend/project
+         seashell/backend/files
+;         seashell/backend/template
```
e45lee: Get rid of unused imports.

src/collects/seashell/backend/backup.rkt
```diff
@@ -0,0 +1,111 @@
...
+  (-> (and/c project-name? is-project?) path-string? void?)
+  (define project-path (build-project-path project))
+  (define backups (if (directory-exists? (get-backup-path project file))
+                      (list-backups project file)
+                      null))
+  (define-values (file-contents file-undo-history) (read-file project file))
+  (define most-recent-backup (if (not (null? backups))
+                               (car backups)
+                               null))
+  (define-values (backup-contents backup-history)
```
e45lee: It's probably easier to just scrub the history when restoring from a backup.

PR #565, opened 2016-08-05
Offline history

mbardakov:
Serialized history for offline mode


e45lee:
Can you make the pull request into offline-compile?  Thanks.

PR #566, opened 2016-08-08
Offline history

mbardakov:
serialized file history for offline mode

### DIFF
--- a/src/collects/seashell/backend/dispatch.rkt
+++ b/src/collects/seashell/backend/dispatch.rkt
@@ -451,7 +451,7 @@
         (_ _) ...)
        (define checksum
          (write-file project file (string->bytes/utf-8 contents)
-                     (if history (string->bytes/utf-8 history) "")
+                     (if history history "")
                      (hash-ref message 'checksum #f)))
        `#hash((id . ,id)
               (success . #t)
@@ -466,7 +466,7 @@
               (success . #t)
               (result .
                       #hash((data . ,(bytes->string/utf-8 data))
-                            (history . ,(bytes->string/utf-8 history))
+                            (history . ,history)
                             (checksum . ,checksum))))]
       ;; Download/Upload token functions:
       [(hash-table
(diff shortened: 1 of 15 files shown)


e45lee:
Can you merge cs136/offline-compile into your branch and resolve the merge conflicts?  Thanks.

src/collects/seashell/backend/dispatch.rkt
```diff
@@ -451,7 +451,7 @@
         (_ _) ...)
        (define checksum
          (write-file project file (string->bytes/utf-8 contents)
-                     (if history (string->bytes/utf-8 history) "")
+                     (if history history "")
```
e45lee: It's probably easier to just leave history as bytes and convert here as your other code assumes
that history is a byte string (snapshots).

mbardakov: I had changed it to a string to get it to work with json->string (or string->json?) somewhere outside my code. Is there somewhere specific I can change it to a string and still leave it as bytes everywhere else?

e45lee: Probably under dispatch.rkt?  (call string->bytes there)

mbardakov: I remember the issue now - json-structs could have strings, but not bytes - so previously I just changed history to be a string *everywhere*. I went back and re-did it so that history is converted to a string for the json-structs, but kept as bytes everywhere else.

src/frontend/js/websocket-service.js
```diff
@@ -378,6 +378,43 @@ angular.module('seashell-websocket', ['ngCookies', 'seashell-local-files', 'seas
         }
       };
 
+
+      self.getMostRecentlyUsed = function(project, directory, deferred) {
```
e45lee: I think these functions are already defined earlier in this file.

e45lee:
Looks good for the most part; just get rid of the functions you defined in websocket-service.js (as I think we already define them).  I will merge this PR once those changes land.

mbardakov:
I didn't actually define anything new here - the way I read the merge
conflict looked like a bunch of stuff from origin and a blank space from
me, so I just kept everything from origin.

e45lee:
OK.  Well, you can safely delete that block now as the offline branch has it's own implementation of the settings/LRU functions.

PR #567, opened 2016-08-11
Syncing project settings and global settings in offline mode

kpalway:


src/backend/compiler/compiler.cc
```diff
@@ -1099,27 +1099,124 @@ extern "C" int seashell_preprocessor_run(struct seashell_preprocessor* preproces
...
+  path.insert(path.begin(), preprocessor->question_dir);
+  path.insert(path.begin(), preprocessor->project_dir);
+
+  // attempt question folder .c file
+  file.pop_back();
+  file.push_back("c");
+  path.pop_back();
+  path.push_back(join(file, '.'));
+  std::string attempt = join(path, '/');
   if(std::ifstream(attempt)) return attempt;
```
e45lee: Probably want to use stat(2) here?  http://man7.org/linux/man-pages/man2/stat.2.html

src/frontend/js/storage-service.js
```diff
@@ -277,9 +293,19 @@
...
+            self.deleteProject(project, true);
+          });
+          updatedProjects.forEach(function(project) {
+            self.updateProject(project);
           });
+          if(settings) {
+            self.saveSettings(JSON.parse(settings));
+          }
           self.database.changelog.clear();
+          Dexie.currentTransaction.on('abort', function(ev) {
```
e45lee: Pull this to the start of the transaction?

lib/libssh2
```diff
@@ -1 +1 @@
-Subproject commit 6d553a7bb94966466c4db0abf9a26660e786a275
+Subproject commit 88366b5ec26c1ac16981e81201cabfb3a6bff8b1
```
e45lee: I think you can resolve this by doing `git checkout  libssh2-1.7.0` from lib/libssh2.

e45lee:
Looks fine, although the segfault is worrisome.  Can you try to fix it?

lib/libssh2
kpalway: Highest version it allows me to checkout here is 1.5.0

kpalway:
The segfault is troubling, but I can't replicate it when I run the tests or do any manual testing. Can somebody pull from this and see if it occurs for them?

kpalway:
Nevermind, the latest commit should fix the problem.

e45lee:
K. Try a 'git fetch' from the libssh2 first to update it? Otherwise, it
looks good to me, and I'll probably merge it tomorrow morning.

kpalway:
Had to change the remote url for libssh2 to the github one, that seemed to fix the problem.

PR #573, opened 2016-09-28
Make files readonly if they contain a magic 'readonly' string

yc2lee:
You can make files readonly by typing these strings anywhere in the file (case-insensitive):

racket: ;; THIS FILE IS READONLY
c/h: // THIS FILE IS READONLY
no extension or txt: THIS FILE IS READONLY

### DIFF
--- a/src/frontend/frontend/file.js
+++ b/src/frontend/frontend/file.js
@@ -18,6 +18,8 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+var SEASHELL_READONLY_STRING = "SEASHELL_READONLY";
+
 /* jshint supernew: true */
 angular.module('frontend-app')
   .controller('EditFileController', ['$state', '$scope', '$timeout', '$q', 'openProject', 'openQuestion',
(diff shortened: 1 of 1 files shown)


kpalway:
I'd prefer if the string said something like SEASHELL_READONLY instead, and if that part of the string were defined as a constant somewhere. Just a minor thing.

dtompkins:
if that's the case, then I'd prefer:

static const int SEASHELL_READONLY = 1;

and

(define SEASHELL_READONLY true)

but I'm fine with

// SEASHELL_READONLY

as well

yc2lee:
I can make the magic readonly string be any sttring, including "(define SEASHELL_READONLY true)", but it'll just be a string equality check... so something like

(define SEASHELL_READONLY (= 1 (+ 0 1)))

won't work... is that okay?

kpalway:
I don't like the idea of adding actual code to the file that isn't intended to be parsed as code.. for me, the comment makes it clearer that this is something Seashell checks on the frontend (and especially not by parsing the code)

PR #574, opened 2016-10-05
Move whitelist projects to different location

kpalway:
Requires passing the SEASHELL_WL_PROJ_BASE_URL parameter to the CMake build


kpalway:
Alright, done

PR #575, opened 2016-10-11
Offline LLVM IR files instead of object files.

kpalway:
Do not merge this yet: most thing seemed to be working until I recompiled the offline compiler. I'm running into a lot of issues with that I'm still trying to resolve.


kpalway:
Building again as RelMinSize seems to make things run again. However the compiler does a console log of "Calling stub instead of sigprocmask()", which I don't recall happening in the past. I'm not sure at this point if this is harmless, or the C runtime isn't built properly.

kpalway:
@e45lee can you look at this and merge some time soon?

PR #584, opened 2016-11-04
Secure whitelist skeleton templates

kpalway:
Uses a system call to SCP to copy the files from the course account instead of downloading them from a publicly readable URL.

### DIFF
--- a/src/collects/seashell/backend/template.rkt
+++ b/src/collects/seashell/backend/template.rkt
@@ -61,8 +61,16 @@
          (match-define (list _ host file) (regexp-match #rx"//(.*@[^:]*):(.*)" source))
          (define-values (sshproc sshout sshin ssherr)
            (subprocess #f #f #f (read-config 'ssh-binary)
-             host (string-append "cat " file)))
-         (thunk sshout)]
+             "-o" "PasswordAuthentication=no" host (string-append "cat " file)))
+         (close-output-port sshin)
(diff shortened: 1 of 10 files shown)


kpalway:
I should have time to change it to use curl

kpalway:
Looks like curl doesn't support sftp:// or scp:// out of the box, so the binary on the linux.student.cs environment won't work with it. Instead I used ssh to clean things up.

I still need to test the behaviour of this when authentication fails, so wait to merge.

src/collects/seashell/backend/template.rkt
```diff
@@ -55,8 +55,14 @@
...
-       [(equal? (url-scheme surl) "file")
-        (call/input-url surl get-pure-port thunk)]
+       [(string=? (url-scheme surl) "file")
+         (call/input-url surl get-pure-port thunk)]
+       [(string=? (url-scheme surl) "ssh")
+         (match-define (list _ host file) (regexp-match #rx"//(.*@[^:]*):(.*)" source))
+         (define-values (sshproc sshout sshin ssherr)
+           (subprocess #f #f #f (read-config 'ssh-binary)
+             host (string-append "cat " file)))
+         (thunk sshout)]
```
e45lee: Make sure you close sshout/ssherr and you use dynamic-wind (see below) to close sshout when thunk returns.

PR #591, opened 2016-11-30
Implement offline testing

yc2lee:
I did some basic testing with tests that pass, fail, compile error, run forever (have to click STOP button), and missing .in or .expect file.

You need my changes in https://github.com/cs136/seashell-clang-js/pull/7 for this to work


kpalway:
It would be ideal if all of the offline/online control logic was pushed down into websocket-service, so that project-service and console.js could treat the offline test results in the same way without any code duplication

PR #590, opened 2016-11-28
Fixes to ASAN parsing and console output buffering (Important)

kpalway:
- Fix segfault ASAN parsing to give stack trace, line numbers
- Add an additional check to the ASAN parser tests
- Make it possible for ASAN parsing code to work from seashell-cli
- Fix frontend console output buffering so we aren't waiting unnecessarily to display program output
- Add some basic tests for seashell-cli

PR #593, opened 2016-12-02
Fix ASAN parser's stack frame regexp match pattern

yc2lee:
A couple more changes to go along with https://github.com/cs136/seashell/pull/590

This also fixes the example program

```
#include <stdio.h>

int g = -1;
int a[] = {0, 1, 2};

int main(void) {
  printf("%d\n", a[g]);
}
```


kpalway:
Was that example program still broken? I thought I had fixed it with my last pull request.

yc2lee:
No I think it was already fixed. I think I probably ran it on the live version

PR #594, opened 2016-12-06
Fixes to offline mode dependency resolution and how stderr is handled offline

kpalway:


src/backend/compiler/compiler.cc
```diff
@@ -990,38 +991,69 @@ extern "C" void seashell_preprocessor_free(struct seashell_preprocessor *preproc
 }
 
 /**
- * seashell_preprocessor_set_main_file(struct seashell_preprocessor* preprocessor, const char* file)
- * Sets the main file to begin running the preprocessor on
+ * seashell_preprocessor_set_question_dir(struct seashell_preprocessor* preprocessor, const char *dir)
```
e45lee: Do we need this for dependency resolution?  I had thought you could pass the preprocessor the same -I flags that we pass to the main clang invocation.

src/backend/compiler/compiler.cc
```diff
@@ -990,38 +991,69 @@ extern "C" void seashell_preprocessor_free(struct seashell_preprocessor *preproc
...
+  preprocessor->question_dir = vec[vec.size()-1];
+  for(int i=0; i<vec.size()-1; i++) {
     preprocessor->project_dir += "/";
     preprocessor->project_dir += vec[i];
   }
-  // add the common folder as a place to look for includes
+  free(nfile);
+
+  // Add directories to look for includes
   preprocessor->compiler_flags.push_back("-I");
```
e45lee: Could you use the main clang flags here also?

src/collects/seashell/compiler/compiler.rkt
```diff
@@ -40,9 +40,10 @@
 
 ;; (seashell-resolve-dependencies file)
 ;; Invokes the internal preprocessor to resolve dependencies
-(: seashell-resolve-dependencies (-> Path (Values (Listof Path) (Listof Seashell-Diagnostic))))
-(define (seashell-resolve-dependencies file)
+(: seashell-resolve-dependencies (-> Path Path (Values (Listof Path) (Listof Seashell-Diagnostic))))
+(define (seashell-resolve-dependencies question-dir file)
```
e45lee: Maybe instead of question-dir pass in the clang flags that we pass to the main clang instance.

src/backend/compiler/compiler.cc
kpalway: I think we need to have both for a case like the following: runner file is common/main.c, It includes common/module.h. The implementation is in the question dir, ie. q1/module.c.

One thing I could do: pass in a list of directory paths which are then added in -I flags and also searched in by the preprocessor for dependency resolution.

e45lee: Sure, that makes sense.  I'm only asking as getting the containing directory for a header file is relatively straightforward with the clang API.

e45lee: see: clang::FileEntry

kpalway: Yeah I remember you mentioning that, but in the case where the header file is in a different directory from its implementation we need to know the question dir anyway (and it can't be retrieved from the #include). And I think it's possible that some assignment question in Seashell might be set up in this way in the future. I can clean up the code a bit by passing the paths and re-using them wherever they're needed.

e45lee: Fair enough.  I'm not entirely sure if we do need to handle that case (as we can simply ship the .h file in the directory where the .c file is supposed to be implemented) but let's find out tomorrow at the meeting.

PR #605, opened 2017-01-06
Build tools to simplify dev installation of seashell. 

kpsuperplane:
New individual scripts in ./build_tools to pull submodules, cmake, and install to ./_build and ./_install. 
`make_install.sh` also automatically sets symlinks and writes "login2.cgi" with the appropriate path. 

Also adds setup.go, which compiles to ./setup, a documented cli tool that calls the above scripts. 

Setup from scratch goes something like the follows:

./setup submodules
./setup cmake
./setup install

Everything is now installed to ./_install and the user only needs to add the etc configuration to get going!


build_tools/cmake.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
e45lee: #!/bin/bash -e -- you want the script to die if any subcommand fails.

build_tools/make_install.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
e45lee: #!/bin/bash -e

build_tools/submodules.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
e45lee: #!/bin/bash -e

build_tools/make_install.sh
```diff
@@ -0,0 +1,16 @@
...
+SCRIPT_PATH=$( cd "$(dirname "${BASH_SOURCE}")" ; pwd -P )
+if [ -d "$SCRIPT_PATH/../_install" ]; then rm -r "$SCRIPT_PATH/../_install/"; fi
+mkdir "$SCRIPT_PATH/../_install/"
+[ -e "$SCRIPT_PATH/../install.log" ] && rm "$SCRIPT_PATH/../install.log"
+echo "Make_installing Seashell to \`/_install\` with 8 cores... logging to /install.log"
+cd "$SCRIPT_PATH/../_build" && make -j8 > ../install.log 2>&1
```
e45lee: Make the -j8 an argument?

build_tools/cmake.sh
```diff
@@ -0,0 +1,8 @@
...
+SCRIPT_PATH=$( cd "$(dirname "${BASH_SOURCE}")" ; pwd -P )
+if [ -d "$SCRIPT_PATH/../_build/" ]; then rm -r "$SCRIPT_PATH/../_build/"; fi
+mkdir "$SCRIPT_PATH/../_build/"
+[ -e "$SCRIPT_PATH/../build.log" ] && rm "$SCRIPT_PATH/../build.log"
+echo "Cmaking Seashell to \`/_build\`... logging to /build.log"
+cd "$SCRIPT_PATH/../_build" && cmake "$SCRIPT_PATH/../" -DCMAKE_INSTALL_PREFIX="$SCRIPT_PATH/../_install"  > ../build.log 2>&1
```
e45lee: 2 comments:
 - You probably want to set LLVM_TARGETS_TO_BUILD=arch/native (can't remember the appropriate one) to speed up the build.
 - Maybe also accept additional arguments to CMake via environment variables/command line.

setup.go
```diff
@@ -0,0 +1,68 @@
+package main
```
e45lee: I'm leery of adding yet another language to the toolchain, especially one that doesn't come in a default Linux installation.

I'd be happier if this script was in Bash (there is a decent way to do command line parsing, see http://wiki.bash-hackers.org/howto/getopts_tutorial), or Python if you need to.

build_tools/make_install.sh
```diff
@@ -0,0 +1,16 @@
...
+cd "$SCRIPT_PATH/../_build" && make install -j8
```
e45lee: Add an optional make test here?

kpsuperplane:
@e45lee I made the changes as requested, and actually removed the go tool completely because I realized the user can simply call ./build_tools/install etc... optional parameters have been added as following:

- ./build_tools/cmake -b "build_path" -i "install_path" -a "extra cmake args" -l "log_path (empty for console)"
- ./build_tools/install -b "build_path" -a "extra make args" -l "log_path (empty for console)" -t "true|false for tests" 

However, it appears travis is failing because my attempt to substitute build arguments into cmake with a variable has made it interpret parts of it as the actual cmake path... any ideas?
image

e45lee:
```
eval cmake "$PROJECT_ROOT" -DLLVM_TARGETS_TO_BUILD="X86" -DCMAKE_INSTALL_PREFIX="$INSTALL_PATH" $BUILD_ARGS
```

e45lee:
Also, for what it's worth, having the setup script would be pretty useful I think, as it's just one script that you need to run to get a "working" Seashell installation.

kpsuperplane:
@e45lee The changes have been made with a quickstart script you can pass -c and -i to to specify cmake and install parameters

PR #626, opened 2017-01-21
Enrich error messages

YLiLarry:
1. display browser info in the error message (browsers tend to lie in userAgent but I couldn't find a better way. at least userAgent gives us a string we can then google for)
2. show a link that has log trace in the error message, recommend user to report the log trace
3. put a reset seashell button in the error message
4. websocket success:false responses are reported with console.error instead of console.log
5. also cleaned up some ugly code. this probably solves the problem where people are getting empty {} error messages.
6. reset seashell? check box is auto selected when appropriate'
7. Add response time to websocket console.log
8. Remove .defer field from websocket console.log
9. websocket request are logged at the time they are sent

### DIFF
--- a/src/frontend/frontend/errors.js
+++ b/src/frontend/frontend/errors.js
@@ -22,28 +22,77 @@ angular.module('frontend-app')
   // Error service.
   .service('error-service', ['$rootScope', '$timeout', '$sce',
     function ($rootScope, $timeout, $sce) {
+      
     var self = this;
+
+    // global variables
     self.errors = [];
+    self.logs = [];
 
-    self.types = {
-      "seashell" : "If this error persists, please email <a href='mailto:seashell@cs.uwaterloo.ca'>seashell@cs.uwaterloo.ca</a> the error message, and the following information for debugging purposes:",
-      "marmoset" : "Make sure you can still access Marmoset's web interface, and try again in a few minutes.",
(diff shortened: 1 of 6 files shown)


src/frontend/frontend/errors.js
```diff
@@ -22,28 +22,83 @@ angular.module('frontend-app')
...
+    // how many entries to keep in self.logs
+    var logSize = 200;
 
-    self.getMessage = function(type) {
-      return $sce.trustAsHtml(type ? self.types[type] : self.types.seashell);
-    };
+    // bug report email format
+    self.reportTo = "seashell@cs.uwaterloo.ca";
+    self.reportSubject = "seashell@cs.uwaterloo.ca";
+    self.reportBody = "Tell us what you were doing when this error showed up. \n[paste the log here]";
```
e45lee: Can you automatically include the logs?

YLiLarry: I'm not sure how to do it.
The reportBody is a parameter in GET url and the logs are too long for that.

e45lee: As far as I can tell, as long as the logs are under <32K, you're good on modern browsers.

YLiLarry:
This is done. Please merge once the test completes.

YLiLarry:
halfcomplexconure-size_restricted
merged yoo!!

PR #628, opened 2017-01-22
move code to fetch new projects to the backend

YLiLarry:
I looked into the frontend code and found some mutual recursive functions different people wrote 
`self.list` and `self.fetch`
and they assumed each other did the actual work

Cleaned up the mess. I moved the logic to backend to save some work later in the frontend write 

Removed some no longer used function in the frontend.

This potentially solves #608

PR #630, opened 2017-01-23
Rewrite fetch skel

YLiLarry:
echo pull #628


e45lee:
Let's put this off for now.

PR #633, opened 2017-01-23
console full screen

YLiLarry:
ctr+shift+enter will fullscreen editor/console depending on which one has cursor focus.

quick switch between editor/console:
ctr+shift+left, right

quit: esc


YLiLarry:
so proud of this. merge please 😆

PR #634, opened 2017-01-24
Fix syncing deadlocking issues

kpalway:


src/collects/seashell/backend/lock.rkt
```diff
@@ -17,34 +17,104 @@
...
+  (start-atomic)
+  (define dead-thds (filter thread-dead? (safe-semaphore-threads sem)))
+  (set-safe-semaphore-count! sem (- (safe-semaphore-count sem) (length dead-thds)))
+  (set-safe-semaphore-threads! sem (remove* dead-thds (safe-semaphore-threads sem)))
+  (cond
+    [(< 0 (safe-semaphore-count sem))
+     (set-safe-semaphore-count! sem (- (safe-semaphore-count sem) 1))
+     (set-safe-semaphore-threads! sem (cons (current-thread) (safe-semaphore-threads sem)))
+     (end-atomic)]
+    [else (end-atomic) (sleep 0) (safe-semaphore-wait sem)]))
```
e45lee: This is going to eat CPU time spinning/waiting on the lock.  You probably ought to  have a list of threads that are suspended on the lock and for safe-semaphore-post to randomly wake up a thread that is suspended.

src/collects/seashell/backend/lock.rkt
```diff
@@ -17,34 +17,104 @@
...
+(define (safe-lock-acquire lock)
+  (start-atomic)
+  (define thd (safe-lock-thread lock))
+  (cond
+    [(eq? (current-thread) thd) (end-atomic)]
+    [(or (not thd)
+         (and thd (thread-dead? thd)))
+     (set-safe-lock-thread! lock (current-thread))
+     (end-atomic)]
+    [else (end-atomic) (sleep 0) (safe-lock-acquire lock)]))
```
e45lee: Use a list of threads that are suspended waiting on the lock.

PR #649, opened 2017-02-07
"Offline-first" saving

kpalway:
- When offline mode is enabled, all writes will save to local storage first then try to sync.
- Reads will also by default read from the local storage.
- Remove separate "connected" and "syncing" events which are not necessary with this change
- The local storage is aware of whether or not it has had an initial sync, and if it has not then it will reject requests until it is synced.


e45lee:
This has landed with the new React UI changes, so I'm closing this.

Issue #648, opened 2017-02-06
Console/Editor size problem


YLiLarry:
Seashell seems like it’s hard coded to 1080P regardless of actual window size, so instead of fixing the “above/below” layout (instead of enlarging the console you shrunk the editor, making both unusably small for use during lectures) you have also broken the side-by-side layout.  At projector resolution (1024x768) not even the entire editor pane fits on screen anymore!
 
So here are the issues:
 
In above/below mode, at the font size I use in class, I get 8 rows in the editor and 7 rows in the console…that’s better than the 2 we had before but well short of the “same height as the editor pane” in the actual feature request.  And it shrunk the editor!  That’s very very bad.
 
In side-by-side mode both editor and console panes are sized to fit at 1080P resolution.  They used to be sized to fit in the actual window!
This also affects above/below, since the total amount of scrolling is identical (and on my 1080P monitor, no scrolling).
 
Here is the desired behavior:
 
Side-by-side mode has the editor and console panes be exactly the height they should be so that the main window does not have a scroll bar (like how it was before)
 
Above/below mode does not touch the editor or console heights at all, it just makes them both full width.  This will make it scroll, but the editor will still exactly fit on screen with the UI so this scrolling does not impact editing at all. 
                                 
-Dan

PR #654, opened 2017-02-09
Enlarge editor

YLiLarry:
request #648 

* Shrink the padding of the entire webpage from 5em to 30px
* Remove margin around the editor when the screen is smaller than 1440px in width
* Clear up the editor size calculation code, remove window-resized event which is only used by this piece of code. Use css instead.
* Editor is now 66% window height in vertical split mode, 70% window height in horizontal split mode

### DIFF
--- a/src/frontend/build/js.frontend
+++ b/src/frontend/build/js.frontend
@@ -1,6 +1,5 @@
 # External Libraries
 //code.jquery.com/jquery-2.1.3.min.js
-//cdnjs.cloudflare.com/ajax/libs/jqueryui/1.12.1/jquery-ui.min.js
 //cdnjs.cloudflare.com/ajax/libs/underscore.js/1.8.3/underscore-min.js
 //cdnjs.cloudflare.com/ajax/libs/sprintf/1.0.2/sprintf.min.js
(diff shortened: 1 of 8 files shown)


kpalway:
I would prefer not to merge a change like this until after the frontend rewrite. Unless we have tested it in all supported browsers with different resolutions including the classroom projectors, and have some consensus among instructors that this is the interface they want.

YLiLarry:
Reverted resizing code. I'm going to make a separate PR for resizable UI.
You need to merge this to fix issue #648

kpalway:
Didn't the revert from earlier this week fix that issue? That was why we reverted it.

YLiLarry:
Olga Dan and I had further discussion about this after the meeting. I think this is what Dan wants.

kpalway:
Okay, has it been tested it on the different browsers and on the projector?

YLiLarry:
I want to have this on staging for instructor feedback.

kpalway:
Okay, make sure we test it thoroughly before our next release, I want to cut down on back and forth UI changes we've been having so far this term.

YLiLarry:
Sounds good! :D

PR #660, opened 2017-02-14
sqash drag-resize-editor

YLiLarry:
Include:
* Support drag resize editor in horizontal split mode.

To-do:
* Find a solution for vertical split mode
* Remember editor size when switching files


e45lee:
We're migrating to a new frontend so closing for now.

PR #662, opened 2017-02-16
[backend/frontend] Output test stdout on errored test.

### LINKED ISSUE #656: Display stdout on crashed tests
Submitter: Tim Brecht
THIS IS AN IMPORTANT BUG THAT NEEDS TO BE FIXED.

There is a bug where output being printed in a program is not appearing as output.
This seems especially problematic when an assertion is triggered in the program.

I've seen this myself and a student had a particularly nasty time trying to debug A5
because none of the printf's output we visible only the fact that an assertion failed
despite the assertion failing after quite a number of prints are called.

fflush(stdout) doesn't solve this problem.

It seems that this might be an issue that occurs when output to the console
is captured before being displayed in the Seashell window.

e45lee:
Fixes #656.

### DIFF
--- a/src/frontend/CMakeLists.txt
+++ b/src/frontend/CMakeLists.txt
@@ -81,7 +81,7 @@ html_compile(${CMAKE_CURRENT_SOURCE_DIR}/index.html)
 find_program(JS_LINTER jshint)
 if (JS_LINTER)
   message("JavaScript linter found: ${JS_LINTER}")
-  add_custom_target(_lint ALL COMMAND ${JS_LINTER} --exclude "'js/lib/*'" . WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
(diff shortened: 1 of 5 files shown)


src/frontend/CMakeLists.txt
```diff
@@ -81,7 +81,7 @@ html_compile(${CMAKE_CURRENT_SOURCE_DIR}/index.html)
 find_program(JS_LINTER jshint)
 if (JS_LINTER)
   message("JavaScript linter found: ${JS_LINTER}")
-  add_custom_target(_lint ALL COMMAND ${JS_LINTER} --exclude "'js/lib/*'" . WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+  # add_custom_target(_lint ALL COMMAND ${JS_LINTER} --exclude "'(js/lib|dist|node_modules)/*'" . WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
```
kpalway: Why is this being commented out?

e45lee: It's cause I can't find a decent way for the build script to exclude multiple directories and in light of our frontend cleanup, I thought it wouldn't be missed.

In particular, jslint linting node_modules/* takes forever and it outputs tons of errors.

PR #664, opened 2017-02-17
storage.js

YLiLarry:
Have tests for all functions except the glue code with websocket


src/frontend/package.json
```diff
@@ -12,10 +12,15 @@
   "license": "MIT",
   "dependencies": {
     "@blueprintjs/core": "^1.6.0",
+    "@types/blueimp-md5": "^1.1.30",
```
e45lee: You probably want to put in a dependency for blueimp-md5 and sprintf via npm -- also jquery.

src/frontend/src/helpers/storage.ts
```diff
@@ -0,0 +1,416 @@
...
+  /*
+    * Save a file to local storage.
+    * @param {string} name: project name.
+    * @param {string} file_name: filename.
+    * @param {string} file_content: The contents of the file.
+    * @param {string} file_history: The history of the file.
+    * @param {string || any false value} checksum : The online checksum of the file, false to not update (offline write).
+    */
+  public writeFile(file: string, contents: string, history: string, checksum: string) {
+    var self = this;
```
e45lee: style: get rid of var self = this as we don't need it anymore.

src/frontend/package.json
```diff
@@ -12,10 +12,15 @@
...
+    "@types/jquery": "^2.0.40",
+    "@types/jquery.cookie": "^1.4.28",
```
e45lee: it would be nice if we didn't need jquery -- can you try using js-cookie?  It's a bit more lightweight.
https://github.com/js-cookie/js-cookie

src/frontend/src/helpers/storage.ts
```diff
@@ -0,0 +1,416 @@
...
+    return self.db.transaction('rw', self.db.changelog, self.db.files, function () {
+      if (checksum !== undefined) {
+        return self.db.files.update(key, {
+          contents: contents, 
+          history: history, 
+          checksum: checksum,
+          last_modified: Date.now()
+        }).then(function (result) {
+          if (! result) {
+            throw sprintf("Storage.writeFile: file %s/%s not found!", name, file);
```
e45lee: We shouldn't actually throw strings.  It would be nice if you could create a class for Storage errors and throw that instead.

e45lee:
Also, you should probably merge the changes in from #649.

e45lee:
One more thing -- can you check out if we can use async/await with Dexie?  I remember that you can, and it would simplify a bunch of the code that is written.

src/frontend/.gitignore
```diff
@@ -1,2 +1,4 @@
 .vscode
-node_modules
\ No newline at end of file
+node_modules
+dist
+/index.html
```
YLiLarry: /index.html is automatically generated?

src/frontend/package.json
```diff
@@ -65,10 +70,12 @@
     "eslint-plugin-jsx-a11y": "^2.2.2",
     "eslint-plugin-react": "^6.3.0",
     "extract-text-webpack-plugin": "^1.0.1",
+    "fake-indexeddb": "git+https://github.com/YLiLarry/fakeIndexedDB.git#3bb9ccc67b2f02bdb2f8d027707903f775b953d0",
```
YLiLarry: hotfixed the compound key update problem

YLiLarry:
I need to merge this to kaleb's branch to continue debugging applyChanges. Please merge this to master react asap is there's big issue with my changes.

PR #669, opened 2017-02-28
Rewrite of websocket code in TypeScript

kpalway:
Couple of things still TODO:
- Add cookie saving/loading code
- Merge with Larry's storage change
- Flesh out some of the placeholder types used (probably goes along with rewriting things like project-service.js)


src/frontend/src/helpers/websocket_client.ts
```diff
@@ -0,0 +1,469 @@
...
+
+  private async authenticate(success: boolean, server_challenge: AuthResult): Promise<boolean> {
+    // Generate a nonce
+    var client_nonce = new Uint8Array(sjcl.random.randomWords(32));
+    for(var i=0; i<client_nonce.length; i++) {
+      client_nonce[i] = client_nonce[i] & 0xFF;
+    }
+
+    // OK, now we proceed to authenticate.
+    var raw_response = new Uint8Array(client_nonce.byteLength + server_challenge.byteLength);
```
e45lee: Look at helpers/crypto.ts for the new authentication interface.

src/frontend/src/helpers/websocket_client.ts
```diff
@@ -0,0 +1,469 @@
...
+    });
+  }
+
+  public close(): void {
+    this.websocket.close();
+  }
+
+  private async authenticate(success: boolean, server_challenge: AuthResult): Promise<boolean> {
+    // Generate a nonce
+    var client_nonce = new Uint8Array(sjcl.random.randomWords(32));
```
e45lee: window.getRandomBytes (so we don't have to depend on another JavaScript library)

src/frontend/src/helpers/websocket_client.ts
```diff
@@ -0,0 +1,469 @@
+import {Coder} from './crypto';
+import * as sjcl from 'sjcl';
+import {History,Settings,Change} from './types';
+
+class Message {
```
e45lee: Easier thing to do here is to use tagged unions -- https://www.typescriptlang.org/docs/handbook/advanced-types.html

src/frontend/src/helpers/websocket_client.ts
```diff
@@ -0,0 +1,469 @@
...
+
+    this.ready = new Promise((resolve, reject)=>{
+      this.websocket = new WebSocket(this.uri);
+      this.started = true;
+      this.websocket.onerror = ()=>{
+        this.failed = true;
+        if(!this.authenticated) {
+          reject("Socket closed during authentication!");
+        }
+        return this.closes && this.closes();
```
e45lee: Errors should be propagated to all waiting promises.

src/frontend/src/helpers/websocket_client.ts
```diff
@@ -0,0 +1,469 @@
...
+
+  /** Sends a message along the connection, ensuring that
+   *  the server and client are properly authenticated.
+   *
+   *  If the socket has not been properly authenticated,
+   *  sends the message after the socket has been properly
+   *  authenticated/set up. */
+  private async sendMessage(message: Message): Promise<WebsocketResult> {
+    if(this.failed || this.closed || !this.started) {
+      throw "Socket closed or failed!";
```
e45lee: Exception class?

src/frontend/src/helpers/websocket.ts
```diff
@@ -0,0 +1,391 @@
...
+      }
+      if(this.offlineEnabled()) {
+        var init = await this.storage.isInitialized();
+        if(!init && !this.isOffline()) {
+          await this.syncAll();
+        } else if(!init) {
+          throw new WebsocketError("Cannot call "+name+" in offline mode before an initial sync.");
+        }
+        var res = await this.storage[name].apply(this.storage, args);
+      }
```
YLiLarry: Can anyone tell me why this is needed?
```
+      if(this.offlineEnabled()) {
 +        var init = await this.storage.isInitialized();
 +        if(!init && !this.isOffline()) {
 +          await this.syncAll();
 +        } else if(!init) {
 +          throw new WebsocketError("Cannot call "+name+" in offline mode before an initial sync.");
 +        }
 +        var res = await this.storage[name].apply(this.storage, args);
 +      }
```
Does this mean if the offline mode is on "enabled", and storage is dirty, `this.syncAll()` is never going to be called here for any of those `offline_enabled` functions?

e45lee: It wouldn't make sense (say reading a file offline when you haven't synced...)

e45lee:
See #687 for this PR.

PR #672, opened 2017-03-06
switch editor over to Ace (https://github.com/ajaxorg/ace), add suppo…

Happytofu:
…rt for Racket colorization, and vim and emacs editor modes


e45lee:
Let's close this PR for now; we can take a look at this again once we have more of the frontend hooked up.

PR #675, opened 2017-03-08
New styles for Settings and Projects

kpsuperplane:
Might break stuff.. @Happytofu @YLiLarry 

screen shot 2017-03-07 at 10 43 44 pm


e45lee:
Two comments for now:
1) How are you going to handle common files and tests?  One thing I could see you doing is that you can have the file tabs named "common/shared_file.c" and "tests/test_1.in" for example.
2) Can we get the tabs list to scroll (if there are too many files, for example).

I'll leave this for @kpalway to look at also when he has more time.

kpsuperplane:
@e45lee yes, scrolling (although not implemented yet) was how I was thinking of doing it. I need to find a place to fit in a submit and add file button...

e45lee:
Sure, that's fine.  You may find it worthwhile to use the newer Tab(s)2 API in blueprint, as the current one is deprecated -- it also has the benefit that you can also stuff other things into the tab bar if you want.

YLiLarry:
Good job making this UI looks very beautiful.
I think this can be done. The goal is to minimize # clicks when you're working on the same file and maximize the editing space.

1. move Test and Run button to the left side bar
2. replace "A1 Racket" by the current question "Q1". You should be able to click on the current question which opens a selection dropdown for other questions "Q2" "Q3..."

YLiLarry:
Good job making this UI looks very beautiful.

I think this can be done. The goal is to minimize # clicks when you're working on the same file and maximize the editing space.

1. move Test and Run button to the left side bar
2. replace "A1 Racket" by the current question "Q1". You should be able to click on the current question which opens a selection dropdown for other questions "Q2" "Q3..."
3. move the file bar to the right side of the current question "Q1"
4. Move/rename Copy Delete ... bar is auto hidden and expanded when the mouse hovers the file bar (on file name)
5. move set Run file to the side bar

Doing this will reduce the top bar to 1, while utilizing the side bar more

YLiLarry:
Honestly I think this is the last thing we should implement.

YLiLarry:
Let's make it work first.

e45lee:
Can I circulate this image to the relevant instructors?  I'd like to land this as soon as possible to prevent any conflicts down the road.

kpsuperplane:
I'm okay with that. Ill take their feedback in conjunction with Larry's and also some of my friends in 136 and get a final draft out after that

kpsuperplane:
Now includes XTerm on the side

kpsuperplane:
Migrated to Tabs2, loading bars are now available

kpsuperplane:
@e45lee do you mind if I merge this. There are a lot of changes in this PR at this point and we can always change the interface later.

djholtby:
It's good there's less wasted whitespace but I echo the concern for fitting common files and tests into that, and wonder how it will work when there are many files and the font size is large.  

Also missing:  Resize button, new file button, new question button.

Please don't have any UI elements that move around just because of where the mouse is hovering.

YLiLarry:
"Please don't have any UI elements that move around just because of where the mouse is hovering."
Sometimes this has to be done to save space. But if we do this, there will be just one tool bar that does this. More than one hiding and expanding UI is confusing, I agree.

e45lee:
I'm fine with the 3 vertical bars.  If you really want vertical space, I suppose the best option is to move all the UI buttons into the sidebar making sure they're aligned with their corresponding elements in the main portion of the screen.

e45lee:
We'll land this for now though.

PR #677, opened 2017-03-09
Console

foleyfactor:
The console should now look right and function more or less correctly.


kpsuperplane:
@e45lee do you have any problems with this? If not I'll merge it so I can adapt my UI revamp to it

e45lee:
Go ahead, but keep in mind that we're going to have to change on('key') to talk with the server later -- for now having the simulated terminal buffer is good.

PR #685, opened 2017-03-11
Update UI based on Professor feedback

kpsuperplane:
image


YLiLarry:
Awesome. Also I think they want to remove the sidebar?

e45lee:
Looks good.  One comment, and this may just be as we've not hooked everything up yet; on initial load, "main.c" is not listed in the tabs list, even though it's open.

kpsuperplane:
Yeah that's a problem with it not being hooked up yet. Merging

PR #687, opened 2017-03-13
Storage websocket

YLiLarry:
for comparison


src/frontend/test/websocket.test.ts
```diff
@@ -0,0 +1,112 @@
...
+        Storage, 
+        File, 
+        Project} from "../src/helpers/websocket";
+import {Login, PublicKey} from "../src/helpers/login";
+import * as R from "ramda";
+import FakeIndexedDB = require("fake-indexeddb");
+import FDBKeyRange = require("fake-indexeddb/lib/FDBKeyRange");
+import {map, filter, flatten, repeat, head, prop} from "ramda";
+import J = require("jscheck");
+import {TestAccount} from "./account";
```
YLiLarry: You need to create a /test/account.ts with the following contents:
```
export {TestAccount}

namespace TestAccount {
  export const user = ""
  export const password = ""
}
```

src/frontend/src/helpers/websocket.ts
```diff
@@ -0,0 +1,411 @@
...
+    this.writeFile = this.make_offline_enabled('writeFile', true);
+    this.deleteFile = this.make_offline_enabled('deleteFile', true);
+    this.renameFile = this.make_offline_enabled('renameFile', true);
+    this.saveSettings = this.make_offline_enabled('saveSettings', true);
+    this.setFileToRun = this.make_offline_enabled('setFileToRun', true);
+    // this.updateMostRecentlyUsed = this.make_offline_enabled('updateMostRecentlyUsed', true)
+  }
+
+  // Connects the socket, sets up the disconnection monitor
+  public async connect(): Promise<void> {
```
YLiLarry: Probably want to design this function so that if user is disconnected, call this function directly to reconnect.

src/frontend/package.json
```diff
@@ -14,16 +14,21 @@
   "license": "MIT",
   "dependencies": {
     "@blueprintjs/core": "1.11.1",
+    "@types/jquery": "2.0.41",
```
YLiLarry: Redundant dependencies

src/frontend/src/helpers/crypto.ts
```diff
@@ -1,3 +1,58 @@
...
+        tmp = arr[i/4];
+      }
+      out.push(tmp >>> 24);
+      tmp <<= 8;
+    }
+    return out;
+  }
+}
+
+/*
```
YLiLarry: I'm having trouble with the encryption.
Somehow the server doesn't accept my response. Even if I just copied the angular code over, the ciphertext is still not accepted by the server.
Somebody more knowledgable about this please help me out.

Happytofu: This is just a guess, but maybe try replacing the second parameter of sjcl.mode.gcm.encrypt with plainArr? Apparently the function's second parameter is supposed to be the plaintext. Again just a wild guess though.

YLiLarry: Yeah the problem is I copied this part from the angular code.. No idea why it does not work. I don't know if it has to do with network configuration. You can checkout my branch, create a file for your user name and password and run `npm test websocket.test.ts`

src/frontend/src/helpers/websocket_client.ts
```diff
@@ -0,0 +1,501 @@
...
+  public async authenticate(): Promise<void> {
+    const server_challenge = await this.requests[-1].received;
+    console.warn(`Server challenge: ${server_challenge}`);
+    /** Generate a nonce. */
+    var client_nonce = sjcl.random.randomWords(32);
+    for (var i = 0; i < client_nonce.length; i++) {
+      client_nonce[i] = client_nonce[i] & 0xFF;
+    }
+
+    /** OK, now we proceed to authenticate. */
```
YLiLarry: So as for the authentication failure, somewhere here is wrong...

src/frontend/src/helpers/crypto.ts
e45lee: Keep in mind that you should use the new crypto code.

Happytofu: What is frame supposed to be anyway?

e45lee: It's supposed to be what you want to encrypt.

Happytofu: Then what's plain?

e45lee: Extra authenticated data that can be sent along in plain-text.

e45lee: As far as I can tell, there's something messed up with SJCL and how it interacts with our test environment here, as it's producing incorrect output.  You should dump SJCL anyways to use the browser's native crypto functions.

e45lee: Actually, your real problem here is that you're passing a TypedArray instance to SJCL for the AES key, which causes a silent failure somewhere.

I've pushed a test to my branch -- e45lee/storage-websocket, but as I said, don't use SJCL.

YLiLarry:
Close for #701

PR #689, opened 2017-03-13
Allow viewing Marmoset test results using websocket - backend

yc2lee:
This is just the backend part.


e45lee:
Do you want to commit the test fetcher tool also?  Something like that would go into src/helpers.  It just helps with keeping track of dependencies for Seashell.

Make sure you sanitize the tool (remove passwords/secrets) though before you commit.

Otherwise, this looks good.

yc2lee:
okay I committed the fetcher tool

e45lee:
LGTM - I'll merge tomorrow.

PR #691, opened 2017-03-13
Fixed terminal scrolling issue.

foleyfactor:
This fixes the bug that @kpsuperplane was getting with the console breaking when receiving data when it needed to scroll.


YLiLarry:
Hey if the change is one or two lines and you're sure there's no problem, just merge it.

YLiLarry:
@kpsuperplane You two check each others work. If it's a bug fix just merge.

PR #693, opened 2017-03-13
Mobile support

kpsuperplane:


kpsuperplane:
Added better dark themeing and settings defaults

kpsuperplane:
@e45lee  not much I can do about getting cut off that's blueprint being blueprint. It was never intended to be mobile.

e45lee:
No problem -- go ahead and merge when you're ready.

PR #701, opened 2017-03-17
Storage + Websocket

YLiLarry:
Use this to continue your frontend developement. 

How to use
* `Import {Storage} from "helpers/Storage.ts"`. This is your entry point. You shouldn't need any file in the `helpers/Storage/*`. If there is any, please add and re-export from `helpers/Storage.ts` instead of including from subdirectories. (In case of internal changes we don't break anything outside the directory.)

* To get a `Connection` object, `import "helpers/Login"`
* You can safely use any function in `helpers/Storage/Interface.ts`
* Please refer to `ProjectID` and `FileID` instead of file and project names. We will need to change the backend to use integer ids later. Current File id is `[string, string]`, so you'll have to use `R.equals` to compare. `==` only compares addresses.

What's tested:
* All functions in `helpers/Storage/Interface.ts`
* Websocket connection with the live server (any change will affect your `https://www.student.cs.uwaterloo.ca/~yourid/cs136/seashell`)
* Local storage functions

To do:
* Apply changes when switching between offline and online mode
* Get rid of sjcl and jquery

Please squash merge.


src/frontend/src/helpers/Login.ts
```diff
@@ -0,0 +1,59 @@
...
+class LoginError extends Error {
+  constructor(msg: string,
+              public username?: string,
+              public status?: number,
+              public statusText?: number) {
+    super(msg);
+  }
+}
+
+class Login {
```
e45lee: Not sure this needs to be class -- wouldn't login(url, user, password, string, reset) be easier?

src/frontend/src/helpers/Login.ts
```diff
@@ -0,0 +1,59 @@
...
+      const msg        = ajax.responseJSON.error.message;
+      const status     = ajax.status;
+      const statusText = ajax.statusText;
+      throw new LoginError(`Login failure (${code}): ${msg}`, user, status, statusText);
+    }
+  }
+}
+
+
+class Connection {
```
e45lee: Maybe store the other information that's relevant -- key, host, port, etc.

src/frontend/src/helpers/Storage/Interface.ts
```diff
@@ -0,0 +1,80 @@
...
+  public abstract async getRunFile(proj: ProjectID, question: string): Promise<FileID|undefined>;
+  public abstract async renameFile(file: FileID, newName: string): Promise<void>;
+  public abstract async deleteFile(file: FileID): Promise<void>;
+  // settings
+  public abstract async setSettings(settings: Settings): Promise<void>;
+  public abstract async getSettings(): Promise<Settings>;
+}
+
+
+type FileID = [string, string]; // compound key
```
e45lee: FileID should be String in the future.

src/frontend/src/helpers/Storage/LocalStorage.ts
```diff
@@ -91,12 +52,6 @@ class Store implements StorageInterface {
   // Will be replaced by Dexie.Syncable.ISyncProtocol
   public async getProjectsForSync(): Promise<Project[]> {
     return this.db.projects.toArray();
```
e45lee: See #681 -- you will need to apply the changes there here.

src/frontend/src/helpers/Storage/Interface.ts
```diff
@@ -0,0 +1,80 @@
...
+type ProjectID = string; // alias of name for now
+
+interface File extends FileBrief {
+  id: FileID;
+  name: string; // a file name is (test|q*|common)/name
+  project: ProjectID;
+  contents: string;
+  checksum: string;
+  lastModified: number;
```
e45lee: last_modified -- don't change the schema.

src/frontend/src/helpers/Storage/LocalStorage.ts
```diff
@@ -153,104 +108,108 @@ class Store implements StorageInterface {
...
       }
-      await this.deleteFile(proj, name, !!checksum);
-      return result.checksum;
+      await this.db.changelog.add(change);
+      await this.deleteFile(id);
     });
   }
 
-  public async getFileToRun(proj: string, question: string): Promise<string|undefined> {
+  public async getRunFile(proj: ProjectID, question: string): Promise<FileID|undefined> {
```
e45lee: Avoid changing the interface type.

src/frontend/src/helpers/Storage/LocalStorage.ts
```diff
@@ -153,104 +108,108 @@ class Store implements StorageInterface {
...
     let p = await this.db.projects.get(proj);
     return p.runs[question];
   }
 
   // a file name is (test|q*|common)/name
-  public async setFileToRun(proj: string, question: string, file: string): Promise<void> {
+  public async setRunFile(file: FileID, question: string): Promise<void> {
```
e45lee: Avoid changing the contract -- Also, you still _need_ to pass in the project, as in the future FileID won't store the project information.

src/frontend/src/helpers/Storage/Interface.ts
```diff
@@ -0,0 +1,80 @@
+export {AbstractStorage,
+        File, FileID, FileBrief,
+        Project, ProjectID, ProjectBrief,
+        Settings, defaultSettings}
+
+abstract class AbstractStorage {
```
e45lee: In the future, the FileID won't hold the project information, so it would be good if we went back to the original contracts and passed along the ProjectID.

src/frontend/src/helpers/Storage/WebStorage.ts
```diff
@@ -0,0 +1,588 @@
...
+  public setOfflineModeSetting(setting: number): void {
+    var old = this.offlineMode;
+    if (setting === 0 || setting === 1 || setting === 2) {
+      this.offlineMode = setting;
+      // Set cookie, set expiry date to some date sufficiently in the future.
+      var expiryDate = new Date();
+      expiryDate.setFullYear(expiryDate.getFullYear() + 10);
+      /* TODO set cookie here
+      cookies.put(SEASHELL_OFFLINE_MODE_COOKIE, setting,
+        {secure: true, expires: expiryDate});
```
e45lee: LocalStorage is probably nicer to use here.

src/frontend/src/helpers/types.ts
```diff
@@ -0,0 +1,48 @@
...
+//   public async getProjectsForSync(): Promise<Array<string>> { return []; }
+//   public async listAllProjectsForSync(): Promise<Array<string>> { return []; }
+//   public async getOfflineChanges(): Promise<Array<Change>> { return []; }
+//   public applyChanges(change: Array<Change>, newProjects: Array<string>, deletedProjects: Array<string>, updatedProjects: Array<string>, settings: Settings) { }
+//   public hasOfflineChanges(): boolean { return false; }
+//   public getSettings(opt?: boolean): any { return {}; }
+// }
+
+export class SeashellCompiler {
+  public async compile(project: string, question: string, files: Array<Object>, runner: string): Promise<any> { return {}; }
```
e45lee: These three classes are basically classes wrapping static functions.  I'd personally just use bare functions (not wrapped in a class), but YMMV.

e45lee:
With switching between offline/online, that's just calling sync...

src/frontend/src/helpers/Login.ts
YLiLarry: Do you mean I should put it in a namespace or make a static class? Or just a global function
Maybe the class name is poorly named. There should also be a `logout` method in there.

src/frontend/src/helpers/Login.ts
YLiLarry: Yeah this is the class that can be stored/retrieved in cookies.
I think the constructor arguments is a shorthand for 
```
public username;
public key;
...
this.username = username;
this.key = key;
...
```

src/frontend/src/helpers/Storage/Interface.ts
YLiLarry: I had the impression that we are moving to replace the sync with dexie's native support. In which case the projectID shouldn't matter. You just need the FileID to update the entry in the file table.

src/frontend/src/helpers/Storage/Interface.ts
YLiLarry: Yeah I can change this.
Do I need to use snake cases for all database entry structures?

src/frontend/src/helpers/Storage/LocalStorage.ts
YLiLarry: The chrome bug? Ok. This part isn't done yet. I'm still working on it :)

src/frontend/src/helpers/Storage/LocalStorage.ts
YLiLarry: I was wavering between the two. I thought if I have FileID I can lookup which table it is in.
If I pass in ProjectID it's also cool.

src/frontend/src/helpers/Storage/WebStorage.ts
YLiLarry: This part is not done.  ;p

src/frontend/src/helpers/types.ts
YLiLarry: I haven't changed this part. These will be removed.

src/frontend/src/helpers/Storage/WebStorage.ts
```diff
@@ -0,0 +1,588 @@
...
+class Callback {
+  constructor(public type: string, public cb: (message?: any) => any, public now: boolean) { }
+}
+
+class WebStorage extends AbstractStorage {
+  private connection: Connection;
+  // private synced: boolean;
+  private connected: boolean;
+  private failed: boolean;
+  // private isSyncing: boolean;
```
kpalway: We need the isSyncing variable or something similar to provide a visual cue when syncing

src/frontend/src/helpers/Storage/WebStorage.ts
```diff
@@ -0,0 +1,588 @@
...
+    // this.writeFile = this.make_offline_enabled('writeFile', true);
+    // this.deleteFile = this.make_offline_enabled('deleteFile', true);
+    // this.renameFile = this.make_offline_enabled('renameFile', true);
+    // this.setSettings = this.make_offline_enabled('setSettings', true);
+    // this.updateMostRecentlyUsed = this.make_offline_enabled('updateMostRecentlyUsed', true)
+  }
+
+  public async getRunFile(proj: ProjectID, question: string): Promise<FileID> {
+    try {
+      const result = await this.socket.sendMessage({
```
kpalway: I'd rather not directly call sendMessage from this class. But if we do, it makes the interface functions that exist in the websocket class redundant.

src/frontend/src/helpers/Storage/WebStorage.ts
```diff
@@ -0,0 +1,588 @@
...
+      if (!this.isOffline()) {
+        return this.socket[name].apply(this.socket, args);
+      } else {
+        return true;
+      }
+    };
+  }
+
+  private make_offline_enabled(name: string, offlineWriteThrough?: boolean) {
+    const online_arity = this.socket[name].length;
```
kpalway: I suspect you stopped using these functions because they make strict typing difficult. That's fine but we still need to have the logic in the functions we're replacing them with.

src/frontend/src/helpers/Storage/WebsocketClient.ts
```diff
@@ -0,0 +1,365 @@
...
+class WebsocketError extends Error {
+  data: Object;
+  constructor(msg: string, d?: Object) {
+    super(msg);
+    this.data = d;
+  }
+}
+
+interface Message {
+  [index: string]: any;
```
kpalway: It would be nice to have a more specific type

src/frontend/src/helpers/Storage/Interface.ts
```diff
@@ -0,0 +1,80 @@
...
+  public abstract async newProject(name: string): Promise<void>;
+  public abstract async getProject(proj: ProjectID): Promise<Project>;
+  public abstract async getProjects(): Promise<ProjectBrief[]>;
+  public abstract async deleteProject(proj: ProjectID): Promise<void>;
+  public abstract async getFiles(proj: ProjectID): Promise<FileBrief[]>;
+  // files
+  public abstract async newFile(proj: ProjectID, filename: string, contents?: string): Promise<FileID>;
+  public abstract async readFile(file: FileID): Promise<File>;
+  public abstract async writeFile(file: FileID, contents: string): Promise<void>;
+  public abstract async setRunFile(file: FileID, question: string): Promise<void>;
```
kpalway: I'd prefer not to change the names of existing functions

src/frontend/src/helpers/Storage/LocalStorage.ts
```diff
@@ -323,66 +281,61 @@ class Store implements StorageInterface {
...
-    }
-    const checksum = (typeof contents === "string" && md5(contents)) || online_checksum || "";
-    const key = [proj, file];
+  public async newFile(proj: ProjectID,
+                       filename: string,
+                       contents?: string): Promise<FileID> {
+    // // account for base64 encoding
+    // let rmatch: RegExpMatchArray;
+    // if (encoding == "url" &&
+    //     (rmatch = contents.match(/^data:([^;]*)?(?:;(?!base64)([^;]*))?(?:;(base64))?,(.*)/))) {
```
kpalway: We will need to have this logic somewhere to get around the base64 encoding issue when uploading files

src/frontend/src/helpers/Storage/WebStorage.ts
e45lee: Sure no problem.

src/frontend/src/helpers/Storage/LocalStorage.ts
e45lee: Yeah so like just pass in ProjectID -- it makes life easier.

src/frontend/src/helpers/Storage/Interface.ts
e45lee: Yeah I would do that for consistency.  In general you shouldn't change the database interface from what is already live and deployed as that will cause problems.

src/frontend/src/helpers/Storage/Interface.ts
e45lee: So there are times where you need both the FileID and the ProjectID -- for example, setFileToRun or whatever.  Yes, the row for the FileID would also store the ProjectID, but it's way nicer if you don't have to do a lookup.

I think in most cases you don't need the Project though, only for setFileToRun.

src/frontend/src/helpers/Login.ts
e45lee: Well the thing is everything you're dealing with here is global data.  I mean, I don't care if it's just a class with static methods or just global functions, but traditionally we've just used global functions.

src/frontend/src/helpers/Storage/Interface.ts
YLiLarry: Sounds good.

src/frontend/src/helpers/Storage/Interface.ts
YLiLarry: Sounds good!

src/frontend/src/helpers/Storage/LocalStorage.ts
YLiLarry: Sounds good!

src/frontend/src/helpers/Storage/WebStorage.ts
YLiLarry: Yes they do not type check properly. And it loses the ability to extend classes.
I member I also had problem catching errors (can't recall exactly).

Yeah sounds good I will add handlers for the offline logics. This is just a fast merge so that Kevin can work on the frontend.

src/frontend/src/helpers/Storage/WebStorage.ts
YLiLarry: I'm actually thinking of splitting the run code compiler away from Storage. Like make it a separate class or something. And you can call `WebsocketClient.sendMessage` from that class for the compiling IO operation. I think `sendMessage` is high level enough to be used in other classes. Meanwhile I move the rest of interface functions from websocket to WebStorage. Although there won't be many, there will be some discrepancies between the current running websocket APIs and what I need here. For example, for listProject the current API returns a `[string,number,string][]`. I want to convert that to a `File[]`. It gets kind of ugly if I do it in `WebsocketClient`.

I also want to enforce the API interface with the backend. In that case I will make an interface.ts for it. I think WebsocketClient should just worry about handling authentication and sending message.

src/frontend/src/helpers/Storage/WebsocketClient.ts
YLiLarry: Sounds good. Will do that later.

src/frontend/src/helpers/Storage/Interface.ts
YLiLarry: I will change that back.

src/frontend/src/helpers/Storage/LocalStorage.ts
YLiLarry: Can we do this logic in the backend?

src/frontend/src/helpers/Login.ts
YLiLarry: Actually, I remember why I did this..
I wanted the ability to login with other uri other than `https://www.student.cs.uwaterloo.ca/~cs136/seashell/cgi-bin/login2.cgi`,
maybe to `seashell-dev`.

src/frontend/src/helpers/Storage/LocalStorage.ts
kpalway: It is done in the backend call as well, but we need it somewhere in the frontend code for when we're not connected to the websocket

YLiLarry: I see. I'll add it back.

src/frontend/src/helpers/Storage/WebStorage.ts
kpalway: One issue I see with that is we probably only want to have one instance of WebsocketClient, so a separate Compiler class would have to share a WebsocketClient with WebStorage. Which we could do.. but it's not as "nice".

Also it seems like converting from a tuple to a File object could be pretty neatly done in WebsocketClient, that's kind of what I imagined that layer would be for (taking raw response from server and converting to the types we work with on the frontend).

It probably is a good idea to separate the compiling/running code to a seperate interface. Let me work on that next since I was planning on hooking up the offline compiler anyway.

src/frontend/src/helpers/Storage/LocalStorage.ts
```diff
@@ -153,104 +108,108 @@ class Store implements StorageInterface {
...
-      let result = await this.db.files.get(key);
+      let result = await this.db.files.get(id);
       result.name = newName;
       await this.db.files.add(result);
       // also rename in run files
       let dbProj = await this.db.projects.get(proj);
       for (const q in dbProj.runs) {
-        if (dbProj.runs[q] == name) {
-          dbProj.runs[q] = newName;
+        if (dbProj.runs[q] == id) {
```
YLiLarry: bug

src/frontend/src/helpers/Storage/WebStorage.ts
YLiLarry: Sounds good. I'll use WebsocketClient as the layer to convert objects.

src/frontend/src/helpers/Storage/LocalStorage.ts
e45lee: Sure OK.

src/frontend/src/helpers/Login.ts
e45lee: Yeah that's fine, but I would pass that as an argument to login(...).  Your call.

src/frontend/src/helpers/Storage/LocalStorage.ts
e45lee: Yeah these are simple fixes but we _need_ to integrate them before we deploy as otherwise problems.

e45lee:
I'm happy to merge this so long as @kpalway approves also.

YLiLarry:
If you can wait, merge tomorrow afternoon.

src/frontend/src/helpers/Storage/WebStorage.ts
YLiLarry: This part isn't done yet.
Btw, is using an event more appropriate then?

kpalway:
I'm fine with merging so we can keep moving on other things

src/frontend/src/helpers/Storage/Interface.ts
YLiLarry: Shall we just leave it like this? I prefer the naming convention to be consistent.
For the frontend we probably want to commit to a different indexeddb since the current one students use are not working anyways unfortunately. For the backend I think this is the only property that's different. I'm just converting it to snake case when sending message.

src/frontend/src/helpers/Login.ts
YLiLarry: done

src/frontend/src/helpers/Storage/Interface.ts
YLiLarry: done

src/frontend/src/helpers/Storage/Interface.ts
kpalway: Keep it the same as it is for now. Theres nothing wrong with the current schema, not sure what you mean by that.

src/frontend/src/helpers/Storage/WebStorage.ts
kpalway: Probably

src/frontend/src/helpers/Storage/Interface.ts
YLiLarry: "Keep it the same" - so use snake case?

YLiLarry: done

YLiLarry:
I made requested changes. Feel free to merge.

PR #703, opened 2017-03-18
Rearrange WebsocketClient and skeleton for Compiler code

kpalway:
Wanted to get a simple PR out quickly so we don't run into conflicts.

This is the basic skeleton code for how we will separate the compiling code from the storage code. The websocket connection will be a singleton shared by the online version of both the compiler and the storage code.

It might be best to move stuff like the the connect(), disconnect() functions in WebStorage into WebsocketClient.


e45lee:
Merged #704 instead.

PR #704, opened 2017-03-19
Rearrange WebsocketClient and skeleton for Compiler code

YLiLarry:
I realized when we talk about "singleton" what we really mean is a "namespace".
This is an alternative solution to PR #703 
All instances are now global constants. So every reference will be the same instance.
No singleton pattern here.

pasted image at 2017_03_19 02_18 am

How to use:
* use Services.storage() to get a WebStorage
* Call Services.login() before using any Storage interface function. It is safe to call Services.login() multiple times. All instances will remain the same.


src/frontend/src/helpers/Login.ts
```diff
@@ -1,57 +0,0 @@
-import * as $ from "jquery";
```
YLiLarry: This module is replaced by Services.ts

src/frontend/src/helpers/Storage/LocalStorage.ts
```diff
@@ -43,11 +43,13 @@ class LocalStorage implements LocalStorageInterface, AbstractStorage {
...
 
-  public constructor(dbName: string, options?: DBOptions) {
-    this.db = new StorageDB(dbName, options);
+  public constructor(private options?: DBOptions) {
     this.has_offline_changes = false;
   }
 
+  public async connect(dbName: string): Promise<void> {
+    this.db = new StorageDB(dbName, this.options);
+  }
```
YLiLarry: Can only connect after login, which gives you the username.

PR #706, opened 2017-03-19
A couple compiler interface changes and some of the online compiler code

kpalway:
The online compiler stuff I'm still hooking up.

Two reasons for the interface changes: I don't see any need for the react code to be dealing with PID objects at all, and I think the Compiler can decide when to call the startIO() method. The react code should basically be able to call a single method for most of the buttons a user presses.


src/frontend/src/helpers/Compiler/Interface.ts
```diff
@@ -1,38 +1,69 @@
...
 
 export {AbstractCompiler,
         CompilerResult,
         CompilerMessage,
-        Test,
-        PID};
+        TestBrief,
+        Test};
 
 abstract class AbstractCompiler {
```
e45lee: Probably want a is_running() method so that the UI can tell if something's running.

kpsuperplane:
Can we have running events actually

kpalway:
Yeah I was thinking of doing these sort of things as events like Larry mentioned in my last PR. I don't know if there's an event API available or what I have to do to set them up, though.

kpsuperplane:
If you look at the reducers in src/reducers you'll see the sort of API were working with.  {action: action_name, payload: dat}

PR #707, opened 2017-03-21
Display student's program output if timeout or killed. Show reminders…

yc2lee:
This pull request fixes the problem where if the student's program times out or is killed, stdout and stderr output up until that point are NOT displayed to the student. With this pull request, when a timeout or kill action occurs, stdout and stderr contents will be displayed, up to a certain length ('max-output-bytes-to-keep setting, set to 10,000 bytes).

The problem was caused by 2 different things:

1. There wasn't any backend code at all to send output to the front end when timeouts or kill occurs. Adding the code is pretty simple (similar to the "pass" and "fail" cases).
2. If students have an infinite loop that prints output, it will make the internal pipes consume a ton of memory. This causes the ```call-with-limits``` resource limits to trigger in dispatch.rkt. It turns out that in Racket, threads inside ```call-with-limits```'s thunk are subject to the resource limits, but they do NOT trigger the ```exn:fail:resource?``` exception when exceeding those limits. To fix this problem, I make the internal pipes store at most 'subprocess-buffer-size MBs of data (a config setting set to 20MB). Any output from the student's program beyond this 20MB is thrown away (we won't send more than 10,000 bytes to the frontend anyways).

This pull request also has some frontend Javascript code that will print out the message "Still testing. N tests left...." every 8 seconds. I added this because it might be confusing for students to stare at a blank console for 30 seconds waiting for a timeout. I can remove this code if you want. We can add a proper progress-bar-like thing later; this is a quick hack in the meantime.

### DIFF
--- a/src/collects/seashell/backend/runner.rkt
+++ b/src/collects/seashell/backend/runner.rkt
@@ -86,8 +86,8 @@
             (close-output-port raw-stdin)))
 
   ;; Background read stuff.
-  (define-values (buf-stderr cp-stderr) (make-pipe))
-  (define-values (buf-stdout cp-stdout) (make-pipe))
+  (define-values (buf-stderr cp-stderr) (make-pipe (* (read-config-nonnegative-real 'subprocess-buffer-size) 1000000)))
+  (define-values (buf-stdout cp-stdout) (make-pipe (* (read-config-nonnegative-real 'subprocess-buffer-size) 1000000)))
   (define stderr-thread (thread (lambda ()
                                   (logf 'debug "Starting copy port for stderr for program PID ~a." pid)
                                   (copy-port raw-stderr cp-stderr))))
@@ -102,6 +102,14 @@
     (close-output-port out-stdout)
     (custodian-shutdown-all custodian)
     (void))
+
+  ;; Helper function that truncates bytes. Used when students write a ton of output and we
+  ;; don't want to send all of it to the front-end.
+  (: first-n-bytes (->* (Bytes Nonnegative-Real) (Bytes) Bytes))
+  (define (first-n-bytes my-bytes n [too-long-message #"\n... There is more output, but it's not shown ...\n"])
+    (if (> (bytes-length my-bytes) (exact-floor n))
+        (bytes-append (subbytes my-bytes 0 (exact-floor n)) too-long-message)
+        my-bytes))
 
   (define receive-evt (thread-receive-evt))
   (let loop ()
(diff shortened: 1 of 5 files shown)


src/collects/seashell/backend/runner.rkt
```diff
@@ -148,7 +156,14 @@
             (kill-thread stderr-thread)
             (kill-thread stdout-thread)
             (subprocess-kill handle #t)
-            (write (serialize `(,pid ,test-name "timeout")) out-stdout)
+
+            ;; Read stdout, stderr.
+            (close-output-port cp-stderr)
+            (close-output-port cp-stdout)
+            (define stdout (first-n-bytes (port->bytes buf-stdout) (read-config-nonnegative-real 'max-output-bytes-to-keep)))
```
e45lee: (read-bytes ...) will work here.

src/collects/seashell/backend/runner.rkt
```diff
@@ -158,7 +173,14 @@
                     (kill-thread stdout-thread)
                     (set-program-exit-status! pgrm 254)
                     (subprocess-kill handle #t)
-                    (write (serialize `(,pid ,test-name "killed")) out-stdout)
+
+                    ;; Read stdout, stderr.
+                    (close-output-port cp-stderr)
+                    (close-output-port cp-stdout)
+                    (define stdout (first-n-bytes (port->bytes buf-stdout) (read-config-nonnegative-real 'max-output-bytes-to-keep)))
```
e45lee: (read-bytes ...) will work here.

src/collects/seashell/backend/runner.rkt
```diff
@@ -103,6 +103,14 @@
     (custodian-shutdown-all custodian)
     (void))
 
+  ;; Helper function that truncates bytes. Used when students write a ton of output and we
+  ;; don't want to send all of it to the front-end.
+  (: first-n-bytes (->* (Bytes Nonnegative-Real) (Bytes) Bytes))
```
e45lee: Probably don't need this -- read-bytes handles partial reads decently.

yc2lee:
Are you sure because my first-n-bytes also appends a message iff bytes are deleted. If I used (read-bytes ...) I still need to compare the length or something similar to check if I should append the "bytes were deleted" message.

yc2lee:
Should I rename the function and/or implement it using (read-bytes ...) instead?

e45lee:
So the thing is that you don't want to extract everything using (port->bytes).  You probably want (first-n-bytes) to use read-bytes to get the first n bytes, and then use byte-ready? to test if there are remaining bytes.

yc2lee:
Oh okay I see

yc2lee:
okay I think i fixed it!

PR #708, opened 2017-03-21
Prevent students from using offline mode. This was requested by CS136…

yc2lee:
In the CS136 course meeting on March 17, 2017, the CS136 instructors agreed that they would like the radio buttons for using offline mode to be disabled. This pull request does this.


kpalway:
Hold on a bit. This won't stop students from using offline mode if it was already enabled. In fact it forces offline mode...?

kpalway:
Oh I see they can still disable it. But they can still use it?

PR #722, opened 2017-03-24
Fix login

YLiLarry:
A websocket patch that enables you to login from browser


src/frontend/src/helpers/Storage/WebStorage.ts
```diff
@@ -289,9 +288,9 @@ class WebStorage extends AbstractStorage {
...
       // trigger reconnect and sync
-      if (!this.socket.isConnected())
-        this.socket.connect();
-      else {
+      if (!this.socket.isConnected()) {
+        // this.socket.connect();
+      } else {
         this.syncAll();
       }
     }
```
YLiLarry: Don't worry about this part. Not done.

PR #725, opened 2017-03-28
Hook up most of online I/O callbacks

kpalway:
Not all of the cases are implemented yet but the most common stuff is there, and it shows the general structure of how we're implementing this. The idea I have in mind is: there are two action types that the compiler can send to the dispatch, "output" and "programDone". Output's payload is simply a string, programDone has a null payload, just meant to notify when the program is no longer running. All of the other processing of output will be done by the Compiler class (I may consider abstracting the processing of output into a separate class). The output sent to the dispatch function will be the final data that should be printed directly to the console.

### DIFF
--- a/src/frontend/src/helpers/Compiler/OnlineCompiler.ts
+++ b/src/frontend/src/helpers/Compiler/OnlineCompiler.ts
@@ -105,7 +105,7 @@ class OnlineCompiler extends AbstractCompiler {
   }
 
   protected programDone(pid: number) {
-    if(this.activePIDs.length === 0 || this.activePIDs[0] !== pid) {
+    if (this.activePIDs.length === 0 || this.activePIDs[0] !== pid) {
       throw new CompilerError("Program that was not running has ended.");
     } else {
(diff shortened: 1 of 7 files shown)


src/frontend/src/helpers/Compiler/OnlineCompiler.ts
```diff
@@ -100,4 +103,12 @@ class OnlineCompiler extends AbstractCompiler {
       pid: this.activePIDs[0]
     });
   }
+
+  protected programDone(pid: number) {
```
e45lee: You don't call programDone when you're running tests, nor does programDone work when tests are being run.  Maybe I've overlooked something, but where do we clear the list of running PIDs when we're running tests?

kpalway: forgot to add it

kpalway: The commit I just pushed should address this - forgot to make the change earlier

PR #728, opened 2017-04-04
Sync

YLiLarry:
Bad news. 
I have changed FileIDs to strings...
Please advice.


src/frontend/src/helpers/Storage/Interface.ts
```diff
@@ -8,34 +8,42 @@ const defaultSettings: Settings = {
...
   // files
   public abstract async newFile(proj: ProjectID, filename: string, contents?: string): Promise<FileID>;
   public abstract async readFile(file: FileID): Promise<File>;
   public abstract async writeFile(file: FileID, contents: string): Promise<void>;
-  public abstract async setFileToRun(proj: ProjectID, question: string, file: FileID): Promise<void>;
-  public abstract async getFileToRun(proj: ProjectID, question: string): Promise<FileID|undefined>;
   public abstract async renameFile(file: FileID, newName: string): Promise<void>;
   public abstract async deleteFile(file: FileID): Promise<void>;
+  // questions
+  public abstract async setFileToRun(proj: ProjectID, question: string, file: FileID): Promise<void>;
```
e45lee: setFileToRun/getFileToRun shouldn't take file IDs, they should take names as file IDs will refer to a particular revision of a file.

src/frontend/src/helpers/Storage/Interface.ts
```diff
@@ -8,34 +8,42 @@ const defaultSettings: Settings = {
...
+  public abstract async newProject(name: string): Promise<ProjectID>;
   public abstract async getProject(proj: ProjectID): Promise<Project>;
   public abstract async getProjects(): Promise<ProjectBrief[]>;
   public abstract async deleteProject(proj: ProjectID): Promise<void>;
-  public abstract async getFiles(proj: ProjectID): Promise<FileBrief[]>;
+  public abstract async getProjectFiles(proj: ProjectID): Promise<FileBrief[]>;
   // files
   public abstract async newFile(proj: ProjectID, filename: string, contents?: string): Promise<FileID>;
   public abstract async readFile(file: FileID): Promise<File>;
   public abstract async writeFile(file: FileID, contents: string): Promise<void>;
```
e45lee: writeFile should return the next FileID (as writing a file ==> new revision/file ID).  For now you can just return the current fileID.

src/frontend/src/helpers/Storage/Interface.ts
YLiLarry: Actually, I think they can also take IDs. If two files have the same name (conflicts), then I can't decide which file to run for user.

e45lee: The backend will prevent you from compiling if there's a conflict.  Leave as-is.

PR #729, opened 2017-04-11
bugfixes to run file, better spacing for terminal, file splash screen

kpsuperplane:


e45lee:
I will merge this for now, but we should implement LRU list for files/questions.

PR #732, opened 2017-04-12
bug fix

YLiLarry:


src/frontend/src/helpers/Storage/LocalStorage.ts
```diff
@@ -129,26 +139,40 @@ class LocalStorage implements AbstractStorage {
   }
 
   public async getFileToRun(proj: ProjectID, question: string): Promise<FileID|false> {
-    let p = await this.getProject(proj);
-    return p.runs[question] || false;
+    this.debug && console.log(`getFileToRun`);
```
e45lee: Should get/return file _names_.

src/frontend/src/helpers/Storage/LocalStorage.ts
```diff
@@ -193,22 +219,21 @@ class LocalStorage implements AbstractStorage {
     });
   }
 
-  public async newFile(proj: ProjectID,
+  public async newFile(pid: ProjectID,
                        name: string,
-                       contents?: string,
```
e45lee: contents is undefined when a file is unavailable offline (i.e a binary file); note newFile is called in syncAll which has to deal with this.

YLiLarry:
You might have to clear your browser's indexedDB after merge.

src/frontend/src/helpers/Storage/LocalStorage.ts
```diff
@@ -56,9 +59,10 @@ class LocalStorage implements AbstractStorage {
     * @param {string} file_history: The history of the file.
     * @param {string || any false value} checksum : The online checksum of the file, false to not update (offline write).
     */
-  public async writeFile(fid: FileID, contents?: string): Promise<void> {
```
e45lee: content is undefined when a file is unavailable offline -- note that writeFile has to deal with this case as syncAll will be called with binary files.

src/frontend/src/helpers/Storage/LocalStorage.ts
```diff
@@ -269,96 +299,136 @@ class LocalStorage implements AbstractStorage {
...
                             newProjects: string[],
                             deletedProjects: ProjectID[]): Promise<void> {
-    const tbs = [this.db.files, this.db.changeLogs, this.db.projects, this.db.settings];
-    return await this.db.transaction("rw", tbs, async () => {
-      Dexie.currentTransaction.on("abort", () => {
-        console.warn("applyChanges transaction aborted");
-      });
+    this.debug && console.log(`applyChanges`);
+    const tbs = [this.db.files, this.db.projects, this.db.settings, this.db.changeLogs];
+    // return await this.db.transaction("rw", tbs, async () => {
```
e45lee: This needs to be in a _single_ transaction, as otherwise you will run into an inconsistent database.

YLiLarry: Forgot to uncomment it..

kpalway:
Can we also make it so we don't have to clear IndexedDB.. if the schema has changed just make IndexedDB do an update that clears the database.

e45lee:
You can do .version(2) in dexie for dexie to do the autoUpgrade for you I think.

YLiLarry:
Not sure how to do that :( it requires both .version(1) and .version(2) to be defined.

e45lee:
Define both, and the upgrade process should do it for you.

src/frontend/src/helpers/Storage/LocalStorage.ts
YLiLarry: writeFile to binary files?

YLiLarry:
Oh well. It requires me to write an upgrade function. I lost track of what's changed :(

kpalway:
You can just make it clear the database, since that was what you suggested in the first place

e45lee:
Fine -- leave it as is, and we'll deal with it later.  Change the database name though (so instead of seashell-e45lee, do instead seashell2-e45lee).

src/frontend/src/helpers/Storage/LocalStorage.ts
e45lee: Yeah in syncAll.

src/frontend/src/helpers/Storage/LocalStorage.ts
YLiLarry: I understand your reasoning. But what if you want to run each conflict version and see which one runs..?

src/frontend/src/helpers/Storage/LocalStorage.ts
YLiLarry: I don't get this part. How do I know if a file's unavailable offline?

e45lee: writeFile will be called with contents set to undefined in syncAll -- that's the backend letting the frontend know that

1)  the file exists, but
2)  isn't available offline (due to being a binary file, etc...)

src/frontend/src/helpers/Storage/LocalStorage.ts
e45lee: That's more work for the backend, so just save the file name.

src/frontend/src/helpers/Services.ts
```diff
@@ -106,12 +106,20 @@ namespace Services {
 
     // login successful
     await localStorage.connect(`seashell-${connection.username}`);
+    await localStorage.deleteDB();
+    await localStorage.connect(`seashell-${connection.username}`);
```
e45lee: Just use a different name -- don't delete the existing database!

YLiLarry: This is just for development environment. Either you clear the DB manually, or I clear it here. For production I will remove it from here.

e45lee: Yeah but we'll forget about it :P.  Just use a different name -- it'll save everyone some trouble.

src/frontend/src/helpers/Storage/LocalStorage.ts
YLiLarry: Ok this might take some time. I'll keep this on my list. Can I worry about it later since you're in a hurry to go?

e45lee: I don't see the problem here; you can just set contents=undefined, fake the checksum, and the rest of the UI _will_ work.

src/frontend/src/helpers/Services.ts
YLiLarry: like appending a Date.now() after? It will litter your browser with several hundreds megabytes of data

src/frontend/src/helpers/Storage/LocalStorage.ts
YLiLarry: done.

src/frontend/src/helpers/Services.ts
e45lee: No just use seashell2-${connection.username}.  I'll change this myself.

PR #740, opened 2017-04-25
	added some webstorage supports for online only calls

YLiLarry:


e45lee:
Make sure you address the slack comments I gave you before you merge though.

PR #741, opened 2017-04-25
Persistent terminal between assignment questions

foleyfactor:
Make console's contents persist between assignment questions.


src/frontend/src/views/Project/Files/Console/index.tsx
```diff
@@ -45,6 +46,18 @@ export default class Xterm extends React.Component<ConsoleProps, ConsoleState> {
     return document.getElementsByClassName("xterm-rows")[0].childElementCount;
   }
 
+  getText(): string {
+        let rows = document.getElementsByClassName("xterm-rows")[0].children;
```
kpalway: Yikes, there's no way to do this through the API? Could this problem be solved more cleanly by associating the console with the project scope instead of the question scope?

src/frontend/src/views/Project/Files/Console/index.tsx
```diff
@@ -66,6 +79,10 @@ export default class Xterm extends React.Component<ConsoleProps, ConsoleState> {
   componentDidMount() {
     this.term.open(this.container);
     this.setState({ input: true, line: 1, currString: "" });
+    if  (this.props.display.props.appState.currentProject) {
+      let consoleText = this.props.display.props.appState.currentProject.consoleText;
+      if (consoleText) this.term.write(consoleText);
+    }
```
e45lee: If you're going to do this you probably ought to attach this to Redux.

foleyfactor: I'll try to take a look at this tonight and see what I can do.

e45lee: Thanks.  If you can't get around to it, let me know and I'll make the necessary edits.

e45lee: I've removed the file prop from DisplayFile -- I haven't checked if this fixes this PR, but please look if you have a chance.

foleyfactor:
I changed the way I was keeping track of the terminal contents by updating the state during termWrite and termClear. The contents are passed to the console by a prop and written when the component mounts.

PR #743, opened 2017-04-28
add checksum to ChangeLog

YLiLarry:


src/frontend/src/helpers/Storage/LocalStorage.ts
```diff
@@ -424,7 +429,7 @@ class LocalStorage implements AbstractStorage {
         if (change.type === "deleteFile") {
           await this.deleteFile(fid, false);
         } else if (change.type === "editFile") {
-          await this.writeFile(fid, change.contents, false);
+          await this.writeFile(fid, change.contents, undefined, false);
```
e45lee: you want change.checksum here (online checksum from the server)

YLiLarry:
All tests fixed and passed. Could you check again and merge it?
Thanks.

src/frontend/src/helpers/Storage/WebStorage.ts
```diff
@@ -117,16 +114,22 @@ class WebStorage extends AbstractStorage implements AbstractWebStorage {
   };
 
   public async newProject(name: string): Promise<ProjectBrief> {
-    if (this.offlineMode === OfflineMode.On ||
-        this.offlineMode === OfflineMode.Forced) {
-      return this.storage.newProject(name);
-    }
+    // if (this.offlineMode === OfflineMode.On ||
+        // this.offlineMode === OfflineMode.Forced) {
+    const off = this.storage.newProject(name);
```
e45lee: I'd do a sync instead afterwards (as you need to fetch the template files)

src/frontend/src/helpers/Storage/WebStorage.ts
```diff
@@ -417,7 +420,7 @@ class WebStorage extends AbstractStorage implements AbstractWebStorage {
     }
     const changesSent = await this.storage.getChangeLogs();
     const settingsSent = await this.storage.getSettings();
-    if (!fetchServerChanges) {
+    if (! fetchServerChanges) {
       if (changesSent.length === 0) {
```
e45lee: We should also test if we've made any changes to project settings, but that can wait.  The easiest way to do this is to have a localStorage variable that's true if and only if we've made any changes.

PR #746, opened 2017-04-30
fix login and auto reconnection

YLiLarry:


src/frontend/src/actions/index.ts
```diff
@@ -355,8 +355,7 @@ const mapDispatchToProps = (dispatch: Function) => {
             } else {
               const path = window.location.pathname.substring(0,
                 window.location.pathname.lastIndexOf("/"));
-              asyncAction(Services.login(username, password, reset,
-                  PRODUCTION ? `https://${window.location.host}${path}/cgi-bin/login2.cgi` : undefined)).then((response) => {
```
e45lee: Mostly good, but I'd the old code here -- as it lets us test on seashell-dev.student.cs.uwaterloo.ca

src/frontend/test/Services.test.ts
```diff
@@ -34,7 +36,7 @@ if (TestAccount.user) {
 function servicesTest() {
 
   beforeAll(() => {
-    return Services.login(TestAccount.user, TestAccount.password, false, TestAccount.backend);
+    return Services.login(TestAccount.user, TestAccount.password, false, LOGIN_URL);
```
e45lee: I'd still get the details from the JSON file just in case.

PR #749, opened 2017-05-18
SQLite backend interface functions

kpalway:
Doesn't even type check right now, just wondering if I am on the right track at all with the querying.


src/collects/seashell/backend/db-files.rkt
```diff
@@ -0,0 +1,179 @@
...
+         (filter (lambda ([x : JSExpr]) (cast (hash-ref (cast x (HashTable Symbol JSExpr)) 'contents_id) (U String False)))
+                 files))
+    (printf "~s\n" files)
+    (cond
+      [zip?
+        (when (file-exists? target) (delete-file target))
+        (zip-from-dir target tmpdir)]
+      [else
+        (when (directory-exists? target) (delete-directory/files target))
+        (copy-directory/files tmpdir target)])
```
e45lee: Could probably do a rename here instead.

e45lee:
Oops I should've mentioned that racket looks for sqlite3 in /usr/lib/x86_64-..., and that's also hardcoded; what I ended up doing on seashell-dev was manually fetching the deb file for sqlite 3.11 from the 16.04 archive and installing it with dpkg -i.  You can get the deb file here: https://launchpad.net/ubuntu/xenial/amd64/libsqlite3-0

PR #750, opened 2017-05-25
prompt confirmation when deleting project

YLiLarry:


YLiLarry:
I modified based on the master branch. it seems like there are some build errors at the current master branch.

kpalway:
The previous build of master passed though

e45lee:
Seems to me like the TypeScript compiler got stricter about inheritance.  As far as I can tell, this is a small fix (x = function() => public x()) so it would be good if you could fix it.

PR #756, opened 2017-07-05
Use typescript 2.3 and make backend logs quieter when in debug/development mode

yc2lee:
The typescript 2.3 is to fix https://github.com/DefinitelyTyped/DefinitelyTyped/issues/17459
I kept getting an error with typescript v2.2


yc2lee:
pull request containing original backend logging code #632

e45lee:
Looks good to me.  I'm not sure how the original logging code disappeared, but everything checks out.

PR #759, opened 2017-07-07
Sentry log tracking (for frontend only/stable deployment)

cs136:


yc2lee:
I did a couple quick tests by forcing an error to occur by messing up my .seashell folder. Errors are logged properly. I hope that the refresh/code disappearing problem actually generates an error, or it won't get logged.

kpalway:
I don't the refreshing thing generates an error in Seashell.. might just crash the tab from memory use or something

PR #761, opened 2017-07-20
More features for sqlite-backend

kpalway:
- Can revert files to old versions with simple UI
- Less websocket errors thrown
- Function to wait for a sync, we now sync before compiling online
- Garbage collector of old contents, for now it is run whenever the server starts
- Changes sent by the backend propagate through the current react state
- Conflict resolution is done through a prompt with readonly textarea and dropdown select


src/frontend/src/actions/index.ts
```diff
@@ -27,11 +27,20 @@ const mapStoreToProps = (state: globalState) => state;
...
       return result;
     } catch (e) {
+      // in case of a conflict, override regular error handling
+      if (e instanceof ConflictError) {
+        dispatch({
+          type: appStateActions.conflictOccurred,
+          payload: e.results
+        });
+        await actions.dispatch.dialog.toggleResolveConflict();
+        return Promise.reject(e.message);
```
e45lee: throw e.message is equivalent (and we probably don't want to just throw a raw string).

src/frontend/src/views/Project/Prompt/Conflict.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <div className="pt-select pt-fill">
+          <select value={this.state.selectedID} disabled={this.state.disabled} onChange={e => {
+              let sel = this.props.appState.conflictContents.find((c: Contents) => c.id === e.currentTarget.value);
+              this.setState(R.merge(this.state, {selectedID: e.currentTarget.value, selectedContents: sel}));
+            }}>
+            {this.props.appState.conflictContents.map((conts: Contents) =>
+              (<option value={conts.id}>{dateString(conts.time)}</option>))}
+          </select>
+        </div>
+        <textarea className={`pt-input pt-fill ${styles.conflictText}`} readOnly
```
e45lee: Maybe a read only Monaco instance here (so it looks nicer.)

src/frontend/src/helpers/Storage/LocalStorage.ts
```diff
@@ -385,6 +437,10 @@ class StorageDB extends Dexie {
   public projects: Dexie.Table<ProjectStored, ProjectID>;
   public settings: Dexie.Table<SettingsStored, number>;
 
+  // list of [accept, reject] pairs waiting for sync to complete
+  private waitlist: [Function, Function][];
```
e45lee: Use a object type -- {accept: Function, reject: Function}

src/frontend/src/reducers/appStateReducer.ts
```diff
@@ -330,6 +341,74 @@ export default function appStateReducer(state: appStateReducerState = {
...
+            const file = question.currentFile;
+            // update the current file
+            let fchg = chgs.find((chg: any) => chg.table === "files"
+              && chg.key === file.id);
+            if (fchg !== undefined) {
+              // if this file entry has been deleted, just jump out of the file
+              if (fchg.type === S.ChangeType.DELETE) {
+                if (state.currentProject
+                    && state.currentProject.currentQuestion) {
+                  state.currentProject.currentQuestion.currentFile = undefined;
```
e45lee: Maybe a little notice "this file was modified on the server...?"

src/collects/seashell/backend/project.rkt
```diff
@@ -391,17 +401,19 @@
...
+  (define arch-file (build-path (read-config-string 'seashell) "archives"
+    (if archive-name archive-name (string-append (number->string (current-seconds)) ".db"))))
+  (define arch-root (build-path (read-config-string 'seashell) "archives"))
+  (define db-file (build-path (read-config-string 'seashell) (read-config-string 'database-file)))
   (unless (directory-exists? arch-root)
     (make-directory arch-root))
-  (rename-file-or-directory proj-root dir-path)
-  (make-directory proj-root))
+  ;; copy the current database
+  (copy-file db-file arch-file)
```
e45lee: Probably want to hold a write-transaction here (but we can add it later).

PR #765, opened 2017-07-30
[frontend] Added functions to cache credentials.

e45lee:


src/frontend/src/helpers/Websocket/WebsocketClient.ts
```diff
@@ -183,7 +185,7 @@ class SeashellWebsocket {
       }
     };
 
-    this.debug && console.log("Waiting for server response...");
+    console.log("Waiting for server response -- setting timeout for %d seconds...", RESPONSE_TIMEOUT);
```
kpalway: Should probably still have the this.debug condition on these

PR #774, opened 2017-08-03
[frontend] Block first acceptance on applyRemoteChanges.

e45lee:
This changes the sync protocol to block until the server's changes have been successfully applied.

If there are a large number of changes that need to be synced this can block forever.  I've submitted a PR to Dexie that will hopefully mitigate this issue (it does on the one bad example we have right now).

See https://github.com/dfahlander/Dexie.js/pull/570/ and https://github.com/dfahlander/Dexie.js/issues/569


kpalway:
applyRemoteChanges returns a promise?

e45lee:
It does return a promise.  It's not very well documented, but it is in the types definition file: https://github.com/dfahlander/Dexie.js/blob/master/addons/Dexie.Syncable/api.d.ts

PR #779, opened 2017-08-20
fix local tests

YLiLarry:
fake-indexeddb 2.0.3 uses a clone algorithm library call realistic-structured-clone <https://github.com/dumbmatter/realistic-structured-clone> which misses supports for cloning `Class` in Dexie, so I still had to fork my own.

Storage.deleteDB doesn't seem to work.


e45lee:
Looks good, can you just repair the other failing tests though and get rid of deleteDB then?

PR #780, opened 2017-08-21
DebugLogs.ts

YLiLarry:
Added a DebugLogs.ts module.
Logs are put in a separate indexed DB. A window with logs will be open by entering `logs()` in Browsers console.


e45lee:
Can you have it so that the error popups have a link to open a new window with the debug logs?

YLiLarry:
Yes. '

PR #803, opened 2017-09-06
Bugfixes [for project export, resource caching]

e45lee:
Fixes:
 - periodic project exporting (was broken, now works and exports on startup)
 - font caching (Overpass font was pointed at a non-cached URL)


yc2lee:
thanks a lot ed!

PR #827, opened 2017-10-25
Update Dexie and put some external libraries on cs136 websites

yc2lee:


e45lee:
FYI both `dexie-syncable@1.0.0-beta.4` and `dexie-observable@1.0.0-beta.4` contain the fixes you need and they're both available through the NPM repository, which is probably better than storing them in the course account.

Same comment for the seashell-clang-js binaries; the bintray stores the results of the automated TravisCI builds, so you can always fetch the latest build there.

